PR #11, opened 2024-05-28
Add support for loading scan tool configuration files

Xenowa:
## Purpose

Adds support for loading the following scan tool configurations from a `toml` file:
- analyzer plugins - compiler plugins that perform external static code analysis.
- rules - static analysis rule identifiers to filter analysis issues.
- platform plugins - static analysis platforms to report analysis issues.


scan-command/src/main/java/io/ballerina/scan/internal/ProjectAnalyzer.java
```diff
@@ -51,4 +65,51 @@ private Consumer<DocumentId> analyzeDocument(Module module, ScannerContextImpl s
...
+        StringBuilder tomlDependencies = new StringBuilder();
+        int importCounter = 0;
+
+        for (ScanTomlFile.Analyzer analyzer : scanTomlFile.getAnalyzers()) {
+            String reportingSource = analyzer.getOrg() + FORWARD_SLASH + analyzer.getName();
+            String analyzerImport = IMPORT_PREFIX + reportingSource + USE_IMPORT_AS_SERVICE;
+            newImports.append(analyzerImport).append("\n");
+
+            if (analyzer.getVersion() != null) {
+                tomlDependencies.append("\n");
```
poorna2152, line 80: Can we use this instead. Check other places as well
```suggestion
                tomlDependencies.append(System.lineSeparator());
```

scan-command/src/main/java/io/ballerina/scan/internal/ProjectAnalyzer.java
```diff
@@ -51,4 +65,51 @@ private Consumer<DocumentId> analyzeDocument(Module module, ScannerContextImpl s
...
+                tomlDependencies.append("[[dependency]]" + "\n");
+                tomlDependencies.append("org='").append(analyzer.getOrg()).append("'\n");
```
poorna2152, line 82: Shall we extract out `analyzer.getOrg()`, `anaylzer.getVersion()` and `analyzer.getName()` out into variables

scan-command/src/main/java/io/ballerina/scan/internal/ProjectAnalyzer.java
```diff
@@ -51,4 +65,51 @@ private Consumer<DocumentId> analyzeDocument(Module module, ScannerContextImpl s
...
+                tomlDependencies.append("name='").append(analyzer.getName()).append("'\n");
+                tomlDependencies.append("version='").append(analyzer.getVersion()).append("'\n");
```
poorna2152, line 84: May be we can extract this appending to a function so we don't have to append the newline repeatedly

scan-command/src/main/java/io/ballerina/scan/internal/ProjectAnalyzer.java
```diff
@@ -51,4 +65,51 @@ private Consumer<DocumentId> analyzeDocument(Module module, ScannerContextImpl s
...
+
+                if (analyzer.getRepository() != null) {
+                    tomlDependencies.append("repository='").append(analyzer.getRepository()).append("'\n");
+                }
+            }
+            importCounter++;
+        }
+
+        Module defaultModule = project.currentPackage().getDefaultModule();
+        Document mainBAL = defaultModule.document(defaultModule.documentIds().iterator().next());
```
poorna2152, line 94: May be this?
```suggestion
        Document mainBal = defaultModule.document(defaultModule.documentIds().iterator().next());
```

scan-command/src/main/java/io/ballerina/scan/utils/ScanUtils.java
```diff
@@ -276,6 +301,242 @@ private static void unzipReportResources(InputStream source, File target) throws
...
+     *
+     * @param project      Ballerina project
+     * @param outputStream print stream
+     * @return an in-memory representation of the Scan.toml file
+     */
+    public static ScanTomlFile loadScanTomlConfigurations(Project project, PrintStream outputStream) {
+        Path targetDir = project.targetDir();
+
+        if (project.kind().equals(ProjectKind.BUILD_PROJECT)) {
+            if (project.currentPackage().ballerinaToml().isPresent()) {
```
poorna2152, line 315: Doesn't the project kind being BUILD_PROJECT guarantee that there is a ballerinaToml

scan-command/src/main/java/io/ballerina/scan/utils/ScanUtils.java
```diff
@@ -276,6 +301,242 @@ private static void unzipReportResources(InputStream source, File target) throws
...
+     * Returns the {@link ScanTomlFile} representation of a local/remote Scan.toml file.
+     *
+     * @param project      Ballerina project
+     * @param outputStream print stream
+     * @return an in-memory representation of the Scan.toml file
+     */
+    public static ScanTomlFile loadScanTomlConfigurations(Project project, PrintStream outputStream) {
+        Path targetDir = project.targetDir();
+
+        if (project.kind().equals(ProjectKind.BUILD_PROJECT)) {
```
poorna2152, line 314: May be we can do an early return here checking for inverse

scan-command/src/main/java/io/ballerina/scan/utils/ScanUtils.java
```diff
@@ -276,6 +301,242 @@ private static void unzipReportResources(InputStream source, File target) throws
...
+                            return loadScanFile(targetDir, absPath.get(), outputStream);
+                        } else {
+                            try {
+                                URL url = new URL(scanTomlPath);
+                                return loadRemoteScanFile(targetDir, url, outputStream);
+                            } catch (MalformedURLException ex) {
+                                throw new RuntimeException("Failed to retrieve remote configuration file " +
+                                        "with exception: " + ex.getMessage());
+                            }
+                        }
```
poorna2152, line 340: Shall we extract this out to a function that retrieves the file content. Looks better that way IMO

scan-command/src/main/java/io/ballerina/scan/utils/ScanUtils.java
```diff
@@ -276,6 +301,242 @@ private static void unzipReportResources(InputStream source, File target) throws
...
+                    analyzer = new ScanTomlFile.Analyzer(org, name, version, repository);
+                } else {
+                    analyzer = new ScanTomlFile.Analyzer(org, name, version, null);
+                }
+                scanTomlFile.setAnalyzer(analyzer);
+            }
+        });
+
+        Toml rulesTable = scanTomlDocumentContent.getTable(RULES_TABLE).orElse(null);
+        if (rulesTable != null) {
```
heshanpadmasiri, line 482: If you can give the inner logic good name factor it to a separate consumer and use `ifPresent` otherwise use `isPresent`

scan-command/src/main/java/io/ballerina/scan/utils/ScanUtils.java
```diff
@@ -276,6 +301,242 @@ private static void unzipReportResources(InputStream source, File target) throws
...
+            TomlValueNode include = rulesTable.get("include").orElse(null);
+            TomlArrayValueNode includeArray;
```
heshanpadmasiri, line 484: Are you using this `includeArray` outside of the if block? Otherwise you can move it in to the if block

scan-command/src/main/java/io/ballerina/scan/utils/ScanUtils.java
```diff
@@ -276,6 +301,242 @@ private static void unzipReportResources(InputStream source, File target) throws
...
+            if (include != null && include.kind().equals(TomlType.ARRAY)) {
```
heshanpadmasiri, line 485: IMO it is cleaner to chain `filter`, `ifPresent` here
```suggestion
            rulesTable.get("include").filter(include -> include.kind().equals(TomlType.ARRAY)).ifPresent(include -> {
            ....
```

scan-command/src/main/java/io/ballerina/scan/utils/ScanUtils.java
```diff
@@ -276,6 +301,242 @@ private static void unzipReportResources(InputStream source, File target) throws
...
+
+                rulesToInclude.forEach(ruleToInclude -> {
+                    ScanTomlFile.RuleToFilter ruleToFilter = new ScanTomlFile.RuleToFilter(ruleToInclude);
+                    scanTomlFile.setRuleToInclude(ruleToFilter);
+                });
+            }
+
+            TomlValueNode exclude = rulesTable.get("exclude").orElse(null);
+            TomlArrayValueNode excludeArray = null;
+            if (exclude != null && exclude.kind().equals(TomlType.ARRAY)) {
```
heshanpadmasiri, line 500: Same as `include`, Also since the inner logic seems similar unless it is accidental may be factor it to a separate method

scan-command/src/main/java/io/ballerina/scan/internal/ProjectAnalyzer.java
```diff
@@ -51,4 +65,51 @@ private Consumer<DocumentId> analyzeDocument(Module module, ScannerContextImpl s
             analyzer.analyze();
         };
     }
+
+    List<Issue> runExternalAnalyzers(Project project) {
+        StringBuilder newImports = new StringBuilder();
+        StringBuilder tomlDependencies = new StringBuilder();
+        int importCounter = 0;
```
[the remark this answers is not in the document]
Xenowa: Yes, since the size of the import counter is the same as analyzer count will make this change.

This value is used to correct the start and end lines of issues found on the in-memory [import-generated Ballerina file](https://github.com/Xenowa/ballerina-scan-tool/blob/main/ScanCommand/src/main/java/io/ballerina/scan/internal/ProjectAnalyzer.java#L453).

scan-command/src/main/java/io/ballerina/scan/internal/ScanToolConstants.java
```diff
@@ -30,6 +30,7 @@ public class ScanToolConstants {
     static final String BALLERINA_ORG = "ballerina";
     static final String BALLERINAI_ORG = "ballerinai";
     static final String BALLERINAX_ORG = "ballerinax";
+    static final String USE_IMPORT_AS_SERVICE = " as _;";
```
MaryamZi, line 33: I would use

```suggestion
    static final String USE_IMPORT_AS_UNDERSCORE = " as _;";
```

It isn't just service, even in the scanner case, it is not.

scan-command/src/main/java/io/ballerina/scan/internal/ProjectAnalyzer.java
Xenowa: In Ballerina projects, the default module can have file names other than 'main.bal', so will change the variable name to a more generic term.

scan-command/src/main/java/io/ballerina/scan/internal/ProjectAnalyzer.java
```diff
@@ -51,4 +65,51 @@ private Consumer<DocumentId> analyzeDocument(Module module, ScannerContextImpl s
...
+
+        for (ScanTomlFile.Analyzer analyzer : scanTomlFile.getAnalyzers()) {
+            String reportingSource = analyzer.getOrg() + FORWARD_SLASH + analyzer.getName();
+            String analyzerImport = IMPORT_PREFIX + reportingSource + USE_IMPORT_AS_SERVICE;
+            newImports.append(analyzerImport).append("\n");
```
MaryamZi, line 77: Do we need the separate `analyzerImport` variable? Why have a separate concatenation anyway? 

Please check and remove unnecessary variables, unless they help with readability.

scan-command/src/main/java/io/ballerina/scan/internal/ProjectAnalyzer.java
```diff
@@ -51,4 +65,51 @@ private Consumer<DocumentId> analyzeDocument(Module module, ScannerContextImpl s
...
+
+            if (analyzer.getVersion() != null) {
+                tomlDependencies.append("\n");
+                tomlDependencies.append("[[dependency]]" + "\n");
+                tomlDependencies.append("org='").append(analyzer.getOrg()).append("'\n");
+                tomlDependencies.append("name='").append(analyzer.getName()).append("'\n");
```
MaryamZi, line 83: Add with spaces around =

scan-command/src/main/java/io/ballerina/scan/internal/ProjectAnalyzer.java
MaryamZi, line 94: balFile

scan-command/src/main/java/io/ballerina/scan/internal/ProjectAnalyzer.java
```diff
@@ -51,4 +65,51 @@ private Consumer<DocumentId> analyzeDocument(Module module, ScannerContextImpl s
...
+
+        List<ScannerContext> scannerContextList = new ArrayList<>();
+
+        // TODO: Implement the logic to pass scanner contexts to compiler plugins and perform external static analysis.
+
+        List<Issue> externalIssues = new ArrayList<>();
+        for (ScannerContext scannerContext : scannerContextList) {
+            ReporterImpl reporter = (ReporterImpl) scannerContext.getReporter();
+            externalIssues.addAll(reporter.getIssues());
+        }
```
MaryamZi, line 112: Do you need a separate list? Why not populate `externalIssues` directly where you will be (instead of) adding to `scannerContextList`?

scan-command/src/main/java/io/ballerina/scan/utils/ScanUtils.java
```diff
@@ -276,6 +301,242 @@ private static void unzipReportResources(InputStream source, File target) throws
...
+            if (project.currentPackage().ballerinaToml().isPresent()) {
+                BallerinaToml ballerinaToml = project.currentPackage().ballerinaToml().get();
+                TomlDocument ballerinaTomlDocument = ballerinaToml.tomlDocument();
+                Toml ballerinaTomlDocumentContent = ballerinaTomlDocument.toml();
+                Toml scanTable = ballerinaTomlDocumentContent.getTable(SCAN_TABLE).orElse(null);
+
+                if (scanTable != null) {
+                    TomlValueNode configPath = scanTable.get(SCAN_FILE_FIELD).orElse(null);
+
+                    if (configPath != null) {
```
MaryamZi, line 324: Check the inverse and return early. That way we can reduce the indentation for the larger part of the logic.

scan-command/src/main/java/io/ballerina/scan/utils/ScanUtils.java
```diff
@@ -276,6 +301,242 @@ private static void unzipReportResources(InputStream source, File target) throws
...
+        }
+        ScanTomlFile scanTomlFile = new ScanTomlFile();
+
+        List<Toml> platformTables = scanTomlDocumentContent.getTables(PLATFORM_TABLE);
+        platformTables.forEach(platformTable -> {
+            Map<String, Object> properties = platformTable.toMap();
+            String name = !(properties.get("name") instanceof String) ? null :
+                    properties.remove("name").toString();
+            String path = !(properties.get("path") instanceof String) ? null :
+                    properties.remove("path").toString();
```
MaryamZi, line 433: Use constants for name, path, etc.

scan-command/src/main/java/io/ballerina/scan/utils/ScanUtils.java
```diff
@@ -276,6 +301,242 @@ private static void unzipReportResources(InputStream source, File target) throws
...
+
+                if (Files.exists(Path.of(path))) {
+                    ScanTomlFile.Platform platform = new ScanTomlFile.Platform(name, path, properties);
+                    scanTomlFile.setPlatform(platform);
+                }
+            }
+        });
+
+        List<Toml> staticCodeAnalyzerTables = scanTomlDocumentContent.getTables(ANALYZER_TABLE);
+        staticCodeAnalyzerTables.forEach(staticCodeAnalyzer -> {
```
MaryamZi, line 454: Generally, normal for loops may be better since they'll allow continuing early too.

scan-command/src/main/java/io/ballerina/scan/internal/ProjectAnalyzer.java
Xenowa: This issue will be addressed in the next PR

scan-command/src/main/java/io/ballerina/scan/utils/ScanUtils.java
```diff
@@ -276,6 +301,242 @@ private static void unzipReportResources(InputStream source, File target) throws
...
+        List<Toml> platformTables = scanTomlDocumentContent.getTables(PLATFORM_TABLE);
+        platformTables.forEach(platformTable -> {
+            Map<String, Object> properties = platformTable.toMap();
+            String name = !(properties.get("name") instanceof String) ? null :
+                    properties.remove("name").toString();
+            String path = !(properties.get("path") instanceof String) ? null :
+                    properties.remove("path").toString();
+
+            if (name != null && !name.isEmpty() && path != null) {
+                if (!(new File(path).exists())) {
```
ShammiL, line 436: We can combine these two conditions

scan-command/src/main/java/io/ballerina/scan/utils/ScanUtils.java
```diff
@@ -276,6 +301,242 @@ private static void unzipReportResources(InputStream source, File target) throws
...
+
+                    if (configPath != null) {
+                        String scanTomlPath = configPath.toNativeValue().toString();
+                        Optional<Path> path = convertConfigurationPath(scanTomlPath);
+                        Optional<Path> absPath = path.isEmpty() || path.get().isAbsolute() ? path
+                                : Optional.of(project.sourceRoot().resolve(scanTomlPath));
+
+                        if (isLocalFile(absPath)) {
+                            return loadScanFile(targetDir, absPath.get(), outputStream);
+                        } else {
```
gayaldassanayake, line 332: Since we are returning from the above else block, we can remove the `else` part.

``` java
if (isLocalFile(absPath)) {
    return loadScanFile(targetDir, absPath.get(), outputStream);
}
try {
    URL url = new URL(scanTomlPath);
    return loadRemoteScanFile(targetDir, url, outputStream);
} catch (MalformedURLException ex) {
    throw new RuntimeException("Failed to retrieve remote configuration file " +
                                        "with exception: " + ex.getMessage());
}

File: scan-command/src/main/java/io/ballerina/scan/utils/ScanUtils.java [hunk 22]
```diff
@@ -276,6 +301,242 @@ private static void unzipReportResources(InputStream source, File target) throws
...
+                if (Files.exists(scanTomlFilePath)) {
+                    outputStream.println("Loading scan tool configurations from " + scanTomlFilePath + "...");
+                    return loadScanFile(targetDir, scanTomlFilePath, outputStream);
+                }
+                return new ScanTomlFile();
+            }
+            return new ScanTomlFile();
+        }
+        return new ScanTomlFile();
+    }
```
gayaldassanayake, line 356:
```suggestion
            }
        }
        return new ScanTomlFile();
    }
```

gayaldassanayake: Even better if you could check all the negative conditions first and early return, then do the real logic.

scan-command/src/main/java/io/ballerina/scan/utils/ScanUtils.java
```diff
@@ -276,6 +301,242 @@ private static void unzipReportResources(InputStream source, File target) throws
...
+            }
+        });
+
+        List<Toml> staticCodeAnalyzerTables = scanTomlDocumentContent.getTables(ANALYZER_TABLE);
+        staticCodeAnalyzerTables.forEach(staticCodeAnalyzer -> {
+            Map<String, Object> properties = staticCodeAnalyzer.toMap();
+            String org = !(properties.get("org") instanceof String) ? null :
+                    properties.get("org").toString();
+            String name = !(properties.get("name") instanceof String) ? null :
+                    properties.get("name").toString();
```
ShammiL, line 459: We can have constants to repeatedly used values

scan-command/src/main/java/io/ballerina/scan/utils/ScanUtils.java
Xenowa: Since the return type of the ballerinaToml() method is optional, will invert conditions with an early return

scan-command/src/main/java/io/ballerina/scan/utils/ScanUtils.java
```diff
@@ -276,6 +301,242 @@ private static void unzipReportResources(InputStream source, File target) throws
...
+    }
+
+    /**
+     * Checks if the provided file is a local file.
+     *
+     * @param path the path of the file
+     * @return true if the file is a local file, false otherwise
+     */
+    private static boolean isLocalFile(Optional<Path> path) throws InvalidPathException {
+        return path.isPresent() && new File(path.get().toString()).exists();
```
gayaldassanayake, line 379:
```suggestion
        return path.isPresent() && path.toFile().isFile()
```

This will also check if this is a file (your code will be true even for a directory)

scan-command/src/main/java/io/ballerina/scan/internal/ProjectAnalyzer.java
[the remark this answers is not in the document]
Xenowa: After the change, a spot bugs warning is generated as the variable is unused. Will be adding this implementation in the next PR.

scan-command/src/main/java/io/ballerina/scan/utils/ScanUtils.java
```diff
@@ -276,6 +301,242 @@ private static void unzipReportResources(InputStream source, File target) throws
...
+                    try {
+                        URL url = new URL(path);
+                        path = loadRemoteJAR(targetDir, name, url, outputStream);
+                    } catch (MalformedURLException ex) {
+                        throw new RuntimeException("Failed to retrieve remote platform file with exception: "
+                                + ex.getMessage());
+                    }
+                }
+
+                if (Files.exists(Path.of(path))) {
```
gayaldassanayake, line 446: If this is the negation of L436 `!(new File(path).exists())`, We can use `else` here.

scan-command/src/main/java/io/ballerina/scan/utils/ScanUtils.java
```diff
@@ -276,6 +301,242 @@ private static void unzipReportResources(InputStream source, File target) throws
...
+            Object providedVersion = properties.get("version");
+            String version = (providedVersion instanceof String) &&
+                    versionPattern.matcher(providedVersion.toString()).matches() ? providedVersion.toString() : null;
+            String repository = !(properties.get("repository") instanceof String) ? null :
+                    properties.get("repository").toString();
+
+            if (org != null && !org.isEmpty() && name != null && !name.isEmpty()) {
+                ScanTomlFile.Analyzer analyzer;
+
+                if (repository != null && repository.equals(LOCAL_REPOSITORY_NAME) && version != null &&
```
ShammiL, line 471: Since there's no other logic between these two if conditions, you could move the `analyzer` initializing outside the condition and combine the two conditions.

scan-command/src/main/java/io/ballerina/scan/utils/ScanUtils.java
```diff
@@ -276,6 +301,242 @@ private static void unzipReportResources(InputStream source, File target) throws
...
+            FileUtils.copyURLToFile(remoteJarFile, cachedJarPath.toFile());
+            outputStream.println("Downloading remote JAR: " + remoteJarFile);
+            return cachedJarPath.toAbsolutePath().toString();
+        } catch (IOException ex) {
+            throw new RuntimeException("Failed to download remote JAR file with exception: " + ex.getMessage());
+        }
+    }
+
     private ScanUtils() {
     }
```
gayaldassanayake, line 541: Shall we move the private constructor to the top. It will show that the class is a util class to anyone who reads it.

scan-command/src/main/java/io/ballerina/scan/utils/ScanTomlFile.java
```diff
@@ -0,0 +1,136 @@
...
+ */
+
+package io.ballerina.scan.utils;
+
+import java.util.Collections;
+import java.util.LinkedHashSet;
+import java.util.Map;
+import java.util.Set;
+
+public class ScanTomlFile {
```
gayaldassanayake, line 26: Let's have javadoc comments here and to all public methods, constructors

scan-command/src/main/java/io/ballerina/scan/internal/ProjectAnalyzer.java
```diff
@@ -51,4 +65,51 @@ private Consumer<DocumentId> analyzeDocument(Module module, ScannerContextImpl s
...
+        Module defaultModule = project.currentPackage().getDefaultModule();
+        Document mainBAL = defaultModule.document(defaultModule.documentIds().iterator().next());
+        String documentContent = mainBAL.textDocument().toString();
+        mainBAL.modify().withContent(newImports + documentContent).apply();
+
+        BallerinaToml ballerinaToml = project.currentPackage().ballerinaToml().orElse(null);
+        if (ballerinaToml != null) {
+            documentContent = ballerinaToml.tomlDocument().textDocument().toString();
+            ballerinaToml.modify().withContent(documentContent + tomlDependencies).apply();
+        }
```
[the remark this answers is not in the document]
Xenowa: Since only atomic variables can be used within lambda expressions, will be implementing this change with a new variable.

scan-command/src/main/java/io/ballerina/scan/utils/ScanUtils.java
Xenowa: This is not possible, as the program should check for the path to the JAR file only if the name and attributes are provided in the `Scan.toml` file. Furthermore, if the provided path is to a remote JAR, it should try to download it first and then assign the downloaded JARs path.

Will inverse the conditions and do an early return to address this.

scan-command/src/main/java/io/ballerina/scan/utils/ScanUtils.java
Xenowa: This logic is to download the remote JAR is the path is a URL, and assign it the path of the downloaded JAR.

scan-command/src/main/java/io/ballerina/scan/utils/ScanTomlFile.java
```diff
@@ -0,0 +1,112 @@
...
+        public Platform(String name, String path, Map<String, Object> arguments) {
+            this.name = name;
+            this.path = path;
+            this.arguments = Collections.unmodifiableMap(arguments);
+        }
+    }
+
+    public record Analyzer(String org, String name, String version, String repository) { }
+
+    public record RuleToFilter(String id) { }
```
MaryamZi, line 111: Docs?

scan-command/src/main/java/io/ballerina/scan/utils/ScanTomlFile.java
```diff
@@ -0,0 +1,112 @@
...
+        /**
+         * @param name      in-memory representation of platform name
+         * @param path      in-memory representation of platform JAR path
+         * @param arguments in-memory representation of platform arguments
+         */
+        public Platform(String name, String path, Map<String, Object> arguments) {
+            this.name = name;
+            this.path = path;
+            this.arguments = Collections.unmodifiableMap(arguments);
+        }
```
heshanpadmasiri, line 106: Since all you need is to make arguments unmodifiable (ie. other parameters are un-touched) you can just use the compact constructor
```suggestion
        public Platform {
            arguments = Collections.unmodifiableMap(arguments);
        }
```

scan-command/src/main/java/io/ballerina/scan/internal/ProjectAnalyzer.java
```diff
@@ -51,4 +64,61 @@ private Consumer<DocumentId> analyzeDocument(Module module, ScannerContextImpl s
...
+        StringBuilder newImports = new StringBuilder();
+        StringBuilder tomlDependencies = new StringBuilder();
+
+        for (ScanTomlFile.Analyzer analyzer : scanTomlFile.getAnalyzers()) {
+            extractAnalyzerImportsAndDependencies(analyzer, newImports, tomlDependencies);
+        }
+        Module defaultModule = project.currentPackage().getDefaultModule();
+        Document balFile = defaultModule.document(defaultModule.documentIds().iterator().next());
+        String balFileContent = balFile.textDocument().toString();
+        balFile.modify().withContent(newImports + balFileContent).apply();
```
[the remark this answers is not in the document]
Xenowa: It currently crashes the program. The other CLI commands produce the following outcome when there are no Ballerina files in a project `ballerina: package is empty. Please add at least one .bal file.`, but proceeds to run if there are Ballerina files in sub-modules. There are three options in this case:

1. Prevent running scans when there are no default module files.
2. Run the scans but prevent running external analyzers.
3. Run the scans and run the external analyzers by importing them to the first found sub-modules Ballerina file.

What approach would be ideal?

scan-command/src/main/java/io/ballerina/scan/utils/ScanTomlFile.java
heshanpadmasiri: Also typically we add the doc comment at the record definition level not at the constructor [see](https://github.com/ballerina-platform/ballerina-lang/blob/0d2fd413d5425689735d259c403b7fdbf8e8846e/semtypes/src/main/java/io/ballerina/types/FixedLengthArray.java#L24)

heshanpadmasiri (commented):
Minor nits, other than that LGTM

scan-command/src/main/java/io/ballerina/scan/internal/ProjectAnalyzer.java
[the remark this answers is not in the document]
MaryamZi: Behaviour should be similar to other commands. 3 sounds okay to me as long as the scan still runs for all files.

Unrelated to this, can you share a snippet of the imports before and after adding the imports the scan tool generates?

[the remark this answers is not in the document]
Xenowa: If we consider a scan is initiated on a newly created Ballerina project, the default module file will look as follows:
```bal
import ballerina/io;

public function main() {
    io:println("Hello, World!");
}
```

If there is a `Scan.toml` file available with the following configurations:
```toml
[[analyzer]]
org = "exampleOrg"
name = "exampleName"

[[analyzer]]
org = "exampleOrg2"
name = "exampleName2"
```

Then the imports generated to the in-memory default module file will be as follows:
```
import exampleOrg/exampleName as _;
import exampleOrg2/exampleName2 as _;
import ballerina/io;

public function main() {
    io:println("Hello, World!");
}
```

Since these imports are generated in-memory and not in the file directly, after the scanning process there will be no changes to the default module file.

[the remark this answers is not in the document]
MaryamZi: This changes line numbers of existing code, doesn't it? That'll affect things like line numbers for compile-time errors. Ideally we should append at the end of an existing import, on the same line.

heshanpadmasiri (approved):
LGTM

scan-command/src/main/java/io/ballerina/scan/internal/ProjectAnalyzer.java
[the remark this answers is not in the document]
Xenowa: Currently working on a fix for this issue, and will send it in the next PR.

> Behaviour should be similar to other commands. 3 sounds okay to me as long as the scan still runs for all files.
> 
> Unrelated to this, can you share a snippet of the imports before and after adding the imports the scan tool generates?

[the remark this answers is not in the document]
Xenowa: > This changes line numbers of existing code, doesn't it? That'll affect things like line numbers for compile-time errors. Ideally we should append at the end of an existing import, on the same line.

Is it alright to always generate an in-memory ballerina file in the default module and add the imports to run the compiler plugins during package compilation?

ShammiL (approved):
LGTM

scan-command/src/test/java/io/ballerina/scan/utils/ScanUtilsTest.java
```diff
@@ -126,4 +118,112 @@ void testGenerateScanReportToProvidedDirectory() throws IOException {
                 .replace(WINDOWS_LINE_SEPARATOR, LINUX_LINE_SEPARATOR);
         Assert.assertEquals(result, expected);
     }
+
+    @Test(description =
+            "test method for loading configurations from a scan toml file in a Ballerina single file project")
+    void testloadScanTomlConfigurationsForSingleFileProject() {
```
poorna2152, line 124: Are we going to add tests where the Scan.toml is resolved using an URL and also negative test cases, where the Scan.toml is not present and so on here?

Xenowa: For testing the `Scan.toml` file resolved from a URL, a remote configuration file needs to exist. So once this PR is merged will be adding tests for loading remote `Scan.toml` configurations by pointing to the file within this repository.

scan-command/src/main/java/io/ballerina/scan/utils/ScanUtils.java
```diff
@@ -333,13 +333,13 @@ public static ScanTomlFile loadScanTomlConfigurations(Project project, PrintStre
...
             }
-            return new ScanTomlFile();
+            return Optional.of(new ScanTomlFile());
         }
 
         Toml scanTable = ballerinaTomlDocumentContent.getTable(SCAN_TABLE).get();
         if (scanTable.get(SCAN_FILE_FIELD).isEmpty()) {
-            outputStream.println("configPath for Scan.toml is missing!");
-            return new ScanTomlFile();
+            outputStream.println("Failed to load scan tool configurations: " + SCAN_FILE_FIELD + " is missing!");
```
gayaldassanayake, line 341: Let's remove the exclamation point in the output log.

scan-command/src/main/java/io/ballerina/scan/utils/ScanUtils.java
```diff
@@ -452,39 +462,43 @@ private static Consumer<Toml> loadAnalyzers(ScanTomlFile scanTomlFile) {
...
+     * Returns the {@link Boolean} value indicating whether the scan process should be continued.
      *
      * @param scanTomlFile the in-memory representation of the Scan.toml file
      * @param targetDir    the target directory
      * @param outputStream the print stream
+     * @return the boolean value indicating whether the scan process should be continued
      */
-    private static Consumer<Toml> loadPlatforms(ScanTomlFile scanTomlFile, Path targetDir, PrintStream outputStream) {
-        return platformTable -> {
+    private static Boolean loadPlatform(Toml platformTable, ScanTomlFile scanTomlFile, Path targetDir,
```
gayaldassanayake, line 472: let's return the primitive type boolean if possible here.

scan-command/src/main/java/io/ballerina/scan/utils/ScanUtils.java
```diff
@@ -452,39 +462,43 @@ private static Consumer<Toml> loadAnalyzers(ScanTomlFile scanTomlFile) {
...
+                                        PrintStream outputStream) {
             Map<String, Object> properties = platformTable.toMap();
             String name = !(properties.get(PLATFORM_NAME) instanceof String) ? null : properties.remove(PLATFORM_NAME)
                     .toString();
```
gayaldassanayake, line 476: Let's swap the two conditions
``` java
String name = properties.get(PLATFORM_NAME) instanceof String ? properties.remove(PLATFORM_NAME)
                    .toString() : null;
```

gayaldassanayake: Makes it more readable

scan-command/src/main/java/io/ballerina/scan/utils/ScanUtils.java
```diff
@@ -452,39 +462,43 @@ private static Consumer<Toml> loadAnalyzers(ScanTomlFile scanTomlFile) {
...
-            String path = !(properties.get(PLATFORM_PATH) instanceof String) ? null : properties.remove(PLATFORM_PATH)
-                    .toString();
+            Optional<String> path = !(properties.get(PLATFORM_PATH) instanceof String) ? Optional.empty() :
+                    Optional.of(properties.remove(PLATFORM_PATH).toString());
 
-            if (name == null || name.isEmpty() || path == null) {
-                return;
+            if (name == null || name.isEmpty() || path.isEmpty()) {
+                return true;
```
gayaldassanayake, line 481: Why are we making `path` Optional<String>` and `name` a `String`?

scan-command/src/main/java/io/ballerina/scan/utils/ScanUtils.java
```diff
@@ -512,26 +526,28 @@ private static Consumer<TomlValueNode> loadRules(ScanTomlFile scanTomlFile, bool
...
      * @param outputStream  the print stream
-     * @return the path of the downloaded remote JAR file
+     * @return the optional path of the downloaded remote JAR file
      */
-    private static String loadRemoteJAR(Path targetDir, String fileName, URL remoteJarFile, PrintStream outputStream) {
+    private static Optional<String> loadRemoteJAR(Path targetDir, String fileName, URL remoteJarFile,
+                                                  PrintStream outputStream) {
         Path cachedJarPath = targetDir.resolve(fileName + JAR_PREDICATE);
         if (Files.exists(cachedJarPath)) {
             outputStream.println("Loading " + fileName + JAR_PREDICATE + " from cache...");
```
gayaldassanayake, line 541: Let's make all the logs similar in format. Here you have used `...`, perhaps to show that it's a continuing process. However in L546, the message is `Downloading remote JAR: xxxxx`, with no dots appended. IMO we should simply display output logs with no exclamations, dot sequences appended to the end.

scan-command/src/main/java/io/ballerina/scan/utils/ScanUtils.java
poorna2152: Also check the place where this function is called too

scan-command/src/main/java/io/ballerina/scan/utils/ScanUtils.java
```diff
@@ -389,44 +390,53 @@ private static ScanTomlFile loadRemoteScanFile(Path targetDir, String scanTomlPa
...
+            return Optional.empty();
         }
         ScanTomlFile scanTomlFile = new ScanTomlFile();
-        scanTomlDocumentContent.getTables(ANALYZER_TABLE).forEach(loadAnalyzers(scanTomlFile));
-        scanTomlDocumentContent.getTables(PLATFORM_TABLE).forEach(loadPlatforms(scanTomlFile, targetDir, outputStream));
+
+        for (Toml platformTable : scanTomlDocumentContent.getTables(PLATFORM_TABLE)) {
+            Boolean shouldContinue = loadPlatform(platformTable, scanTomlFile, targetDir, outputStream);
+
+            if (!shouldContinue) {
```
poorna2152, line 422: We can inline this function call right?

scan-command/src/test/java/io/ballerina/scan/utils/ScanUtilsTest.java
```diff
@@ -183,27 +179,43 @@ void testloadScanTomlConfigurations() {
...
     }
 
+    @Test(description =
+            "test method for loading configurations from a Scan.toml file with invalid platform configuration")
+    void testloadInvalidPlatformScanTomlConfigurations() throws IOException {
+        Path ballerinaProject = testResources.resolve("test-resources")
+                .resolve("bal-project-with-invalid-platform-config-file");
+        Project project = BuildProject.load(ballerinaProject);
+        System.setProperty("user.dir", ballerinaProject.toString());
+        String result = "";
```
poorna2152, line 189: Define this closer to usage. Check other places as well

scan-command/src/test/java/io/ballerina/scan/utils/ScanUtilsTest.java
```diff
@@ -126,4 +120,174 @@ void testGenerateScanReportToProvidedDirectory() throws IOException {
...
+        Set<ScanTomlFile.Analyzer> analyzers = scanTomlFile.getAnalyzers();
+        List<ScanTomlFile.Analyzer> analyzerList = new ArrayList<>(analyzers);
+        Assert.assertEquals(analyzerList.size(), 4);
+        ScanTomlFile.Analyzer analyzer = analyzerList.get(0);
+        Assert.assertEquals(analyzer.org(), "exampleOrg");
+        Assert.assertEquals(analyzer.name(), "exampleName");
+        analyzer = analyzerList.get(1);
+        Assert.assertEquals(analyzer.org(), "ballerina");
+        Assert.assertEquals(analyzer.name(), "example_module_static_code_analyzer");
+        Assert.assertEquals(analyzer.version(), "0.1.0");
```
poorna2152, line 156: If we are asserting analyzers at multiple places would it make sense to extract these assertions to a function

gayaldassanayake (approved):

poorna2152 (approved):

scan-command/src/main/java/io/ballerina/scan/internal/ScanCmd.java
```diff
@@ -131,13 +133,23 @@ public void execute() {
             return;
         }
 
-        ProjectAnalyzer projectAnalyzer = new ProjectAnalyzer();
-        List<Rule> coreRules = CoreRule.rules();
+        if (ProjectUtils.isProjectEmpty(project.get())) {
+            outputStream.println("ballerina: package is empty. Please add at least one .bal file.");
```
MaryamZi, line 137: Thinking out loud, even for other commands, I'm not entirely sure if this is a valid ask, since adding an empty bal file will make this error go away. We can rethink this separately.

scan-command/src/main/java/io/ballerina/scan/internal/ScanToolConstants.java
```diff
@@ -30,6 +30,8 @@ public class ScanToolConstants {
     static final String BALLERINA_ORG = "ballerina";
     static final String BALLERINAI_ORG = "ballerinai";
     static final String BALLERINAX_ORG = "ballerinax";
+    static final String USE_IMPORT_AS_UNDERSCORE = " as _;";
+    static final String IMPORT_GENERATOR_FILE = "scanFile";
```
MaryamZi, line 34:
```suggestion
    static final String IMPORT_GENERATOR_FILE = "scan_file";
```

MaryamZi (approved):
Please rename the file in the next PR.

scan-command/src/main/java/io/ballerina/scan/internal/ProjectAnalyzer.java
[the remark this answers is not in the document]
Xenowa: Since the generation of imports is now done via a generated in-memory Ballerina file, having this functionality is no longer required.

PR #12, opened 2024-05-31
Add support for publishing scan tool

Xenowa:
## Purpose

Add support for triggering workflows to publish the scan tool to:

- GitHub Packages
- Ballerina Central

### DIFF
--- a/.github/workflows/central_publish.yml
+++ b/.github/workflows/central_publish.yml
@@ -1,0 +1,76 @@
+name: Publish to the Ballerina central
+
+on:
+  workflow_dispatch:
+    inputs:
+      environment:
+        type: choice
+        description: Select environment
+        required: true
+        options:
+          - CENTRAL
+          - DEV CENTRAL
+          - STAGE CENTRAL
+
+jobs:
+  central_publish:
+    name: Run central publish workflow
+    runs-on: ubuntu-latest
+
+    if: github.repository_owner == 'ballerina-platform'
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v3
+
+      - name: Set up JDK 17
+        uses: actions/setup-java@v3
+        with:
+          distribution: 'temurin'
+          java-version: '17.0.7'
+
+      - name: Set Up Ballerina
+        uses: ballerina-platform/setup-ballerina@v1.1.0
+        with:
+          version: latest
+
+      - name: Grant execute permission for gradlew
+        run: chmod +x gradlew
+
+      - name: Build without tests
+        env:
+          packageUser: ${{ github.actor }}
+          packagePAT: ${{ secrets.GITHUB_TOKEN }}
+        run: ./gradlew build -x check -x test
+
+      - name: Ballerina Central Push
+        if: ${{ inputs.environment == 'CENTRAL' }}
+        env:
+          BALLERINA_DEV_CENTRAL: false
+          BALLERINA_STAGE_CENTRAL: false
+          BALLERINA_CENTRAL_ACCESS_TOKEN: ${{ secrets.BALLERINA_CENTRAL_ACCESS_TOKEN }}
+          packageUser: ${{ secrets.BALLERINA_BOT_USERNAME }}
+          packagePAT: ${{ secrets.BALLERINA_BOT_TOKEN }}
+          GITHUB_TOKEN: ${{ secrets.BALLERINA_BOT_TOKEN }}
+        run: ./gradlew release -PpublishToCentral=true
+
+      - name: Ballerina Central Dev Push
+        if: ${{ inputs.environment == 'DEV CENTRAL' }}
+        env:
+          BALLERINA_DEV_CENTRAL: true
+          BALLERINA_STAGE_CENTRAL: false
+          BALLERINA_CENTRAL_ACCESS_TOKEN: ${{ secrets.BALLERINA_CENTRAL_DEV_ACCESS_TOKEN }}
+          packageUser: ${{ secrets.BALLERINA_BOT_USERNAME }}
+          packagePAT: ${{ secrets.BALLERINA_BOT_TOKEN }}
+          GITHUB_TOKEN: ${{ secrets.BALLERINA_BOT_TOKEN }}
+        run: ./gradlew release
+
+      - name: Ballerina Central Stage Push
+        if: ${{ inputs.environment == 'STAGE CENTRAL' }}
+        env:
+          BALLERINA_DEV_CENTRAL: false
+          BALLERINA_STAGE_CENTRAL: true
+          BALLERINA_CENTRAL_ACCESS_TOKEN: ${{ secrets.BALLERINA_CENTRAL_STAGE_ACCESS_TOKEN }}
+          packageUser: ${{ secrets.BALLERINA_BOT_USERNAME }}
+          packagePAT: ${{ secrets.BALLERINA_BOT_TOKEN }}
+          GITHUB_TOKEN: ${{ secrets.BALLERINA_BOT_TOKEN }}
+        run: ./gradlew release
--- a/.github/workflows/publish_package.yml
+++ b/.github/workflows/publish_package.yml
@@ -1,0 +1,40 @@
+name: Publish to GitHub package registry
+
+on:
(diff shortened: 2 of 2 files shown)


.github/workflows/central_publish.yml
ThisaruGuruge, line 1: We usually use the `central-publish.yml` for the workflows used to publish packages to Ballerina Central. This helps when we have centralized workflows and creating pipelines so when we need to release a particular package to the Ballerina central from another workflow, we know exactly what workflow to call in each repo. So let's adhere to that practice and name this file as `central-publish.yml`.

.github/workflows/central_publish.yml
```diff
@@ -0,0 +1,76 @@
...
+          distribution: 'temurin'
+          java-version: '17.0.7'
+
+      - name: Set Up Ballerina
+        uses: ballerina-platform/setup-ballerina@v1.1.0
+        with:
+          version: latest
+
+      - name: Grant execute permission for gradlew
+        run: chmod +x gradlew
```
ThisaruGuruge, line 37: We don't need this step. This is present in some workflow for historical reasons, but now the GitHub actions do have access to execute the Gradle scripts.

.github/workflows/central_publish.yml
```diff
@@ -0,0 +1,76 @@
...
+
+      - name: Build without tests
+        env:
+          packageUser: ${{ github.actor }}
+          packagePAT: ${{ secrets.GITHUB_TOKEN }}
+        run: ./gradlew build -x check -x test
```
ThisaruGuruge, line 43: Any reason to skip tests here? Ideally, before publishing, we should run tests and stop the publish task if there are test failures.

.github/workflows/publish_package.yml
ThisaruGuruge, line 1: Let's rename this to `publish-release.yml` for the above mentioned reason.

.github/workflows/publish_package.yml
```diff
@@ -0,0 +1,40 @@
+name: Publish to GitHub package registry
```
ThisaruGuruge, line 1:
```suggestion
name: Release the Package
```

.github/workflows/publish_package.yml
```diff
@@ -0,0 +1,40 @@
...
+          distribution: 'temurin'
+          java-version: '17.0.7'
+
+      - name: Set Up Ballerina
+        uses: ballerina-platform/setup-ballerina@v1.1.0
+        with:
+          version: latest
+
+      - name: Grant execute permission for gradlew
+        run: chmod +x gradlew
```
ThisaruGuruge, line 28: We can remove this.

.github/workflows/central_publish.yml
```diff
@@ -0,0 +1,76 @@
...
+      - name: Ballerina Central Push
+        if: ${{ inputs.environment == 'CENTRAL' }}
+        env:
+          BALLERINA_DEV_CENTRAL: false
+          BALLERINA_STAGE_CENTRAL: false
+          BALLERINA_CENTRAL_ACCESS_TOKEN: ${{ secrets.BALLERINA_CENTRAL_ACCESS_TOKEN }}
+          packageUser: ${{ secrets.BALLERINA_BOT_USERNAME }}
+          packagePAT: ${{ secrets.BALLERINA_BOT_TOKEN }}
+          GITHUB_TOKEN: ${{ secrets.BALLERINA_BOT_TOKEN }}
+        run: ./gradlew release -PpublishToCentral=true
```
ThisaruGuruge, line 54: I am assuming you are using the Ballerina Gradle plugin here. In that case, you don't need the Gradle release task here. You can invoke the `build` task with `-P publishToCentral=true`.

We have two separate workflows for release a package and to publish it to the Ballerina central. Following are the reasons for that:

1. The `publish-release` workflow is the main workflow we use to release a package. This workflow should handle the following:
    i. Create a git tag on GitHub
    ii. Publish the relevant Maven artifacts to GitHub packages
   iii. Publish the Ballerina package to the Ballerina Central
    Additionally, it should send a post-release PR to the default branch after the release to sync the default branch with the release branch. So the `publish-release` workflow should be the place to handle Gradle `release` task.
2. The `central-publish` workflow is mainly used to publish the package to the Ballerina Dev Central and the Ballerina Stage Central. (Additionally is is used to publish the package to the Ballerina Central if the publishing to the Ballerina Central is failed in the `publish-release` workflow due to some intermittent issue).

If we use the Gradle `release` task here, it will remove the `SNAPSHOT` version from the `gradle.properties` file and commit it to the relevant branch, which shouldn't happen in this workflow. So In this case, you only need to use the `-P publishToCentral=true`.

This is true for all three publishing steps in this workflow.

```suggestion
        run: ./gradlew clean build -PpublishToCentral=true
```

.github/workflows/central_publish.yml
```diff
@@ -0,0 +1,76 @@
...
+      - name: Ballerina Central Dev Push
+        if: ${{ inputs.environment == 'DEV CENTRAL' }}
+        env:
+          BALLERINA_DEV_CENTRAL: true
+          BALLERINA_STAGE_CENTRAL: false
+          BALLERINA_CENTRAL_ACCESS_TOKEN: ${{ secrets.BALLERINA_CENTRAL_DEV_ACCESS_TOKEN }}
+          packageUser: ${{ secrets.BALLERINA_BOT_USERNAME }}
+          packagePAT: ${{ secrets.BALLERINA_BOT_TOKEN }}
+          GITHUB_TOKEN: ${{ secrets.BALLERINA_BOT_TOKEN }}
+        run: ./gradlew release
```
ThisaruGuruge, line 65:
```suggestion
        run: ./gradlew clean build -PpublishToCentral=true
```

.github/workflows/central_publish.yml
ThisaruGuruge: As per the comment on line no. 43, we can remove this step altogether since we are using `gradle build` command when publishing, which should invoke the tests as well.

.github/workflows/central_publish.yml
```diff
@@ -0,0 +1,76 @@
...
+      - name: Ballerina Central Stage Push
+        if: ${{ inputs.environment == 'STAGE CENTRAL' }}
+        env:
+          BALLERINA_DEV_CENTRAL: false
+          BALLERINA_STAGE_CENTRAL: true
+          BALLERINA_CENTRAL_ACCESS_TOKEN: ${{ secrets.BALLERINA_CENTRAL_STAGE_ACCESS_TOKEN }}
+          packageUser: ${{ secrets.BALLERINA_BOT_USERNAME }}
+          packagePAT: ${{ secrets.BALLERINA_BOT_TOKEN }}
+          GITHUB_TOKEN: ${{ secrets.BALLERINA_BOT_TOKEN }}
+        run: ./gradlew release
```
ThisaruGuruge, line 76: Ditto

```suggestion
        run: ./gradlew clean build -PpublishToCentral=true
```

.github/workflows/publish_package.yml
```diff
@@ -0,0 +1,40 @@
...
+        env:
+          packageUser: ${{ github.actor }}
+          packagePAT: ${{ secrets.GITHUB_TOKEN }}
+        run: ./gradlew build -x check -x test
+
+      - name: publish package to GitHub package registry
+        env:
+          publishUser: ${{ github.actor }}
+          publishPAT: ${{ secrets.GITHUB_TOKEN }}
+        run: ./gradlew publish
```
ThisaruGuruge, line 40: In this workflow, you can use the Gradle `release` task to release and publish the artifacts. You can check an example [here](https://github.com/ballerina-platform/ballerina-library/blob/803ed204cf711774a2b5e12fde6bb5ceb3c0512b/.github/workflows/release-package-template.yml#L64).

Xenowa:
Closing and moving the configurations of this PR to the `static-code-analysis-tool` branch to have the package publishing logic as a single PR.

.github/workflows/central_publish.yml
Xenowa: Since this is a standalone Ballerina package, the Ballerina Gradle plugin will not be used. Instead, the CLI-based configurations to push the tool will be introduced.

PR #13, opened 2024-06-03
Add support for publishing scan tool

Xenowa:
## Purpose

Add support for triggering workflows to publish the scan tool to:

- GitHub Packages
- Ballerina Central

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,6 @@ build/
 
 # Coverage reports
 .jacoco/
+
+#Ignore all .deb files
+*.deb
--- a/build.gradle
+++ b/build.gradle
@@ -1,3 +1,21 @@
+/*
+ * Copyright (c) 2024, WSO2 LLC. (https://www.wso2.com).
+ *
+ * WSO2 LLC. licenses this file to you under the Apache License,
+ * Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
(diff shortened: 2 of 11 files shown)


build.gradle
```diff
@@ -0,0 +1,20 @@
+plugins {
```
ThisaruGuruge, line 1: Missing license header.

ThisaruGuruge (approved):
LGTM

.gitignore
```diff
@@ -14,3 +14,6 @@ build/
 
 # Coverage reports
 .jacoco/
+
+#Ignore all .deb files
+*.deb
```
keizer619, line 19: @Xenowa Why we had to put this entry? if its created inside build directory, can't we just ignore that?

Xenowa: This entry is needed to address an issue where the GitHub package publishing fails. The failure occurs because the setup Ballerina action downloads a `.deb` file, which is then flagged as uncommitted changes. By explicitly ignoring this `.deb` file we can ensure that this file does not interfere with the publishing process.

keizer619: Can you create an issue in [setup-ballerina](https://github.com/ballerina-platform/setup-ballerina/) repo? Ideally, it should clean everything after setting up and should not keep anything to be committed.

keizer619 (approved):

.gitignore
ThisaruGuruge: We too faced this when we did the connector revamp. We fixed it by adding it to the .gitignore file. Would be great if we can fix this from the setup-ballerina action itself. 
@NipunaRanasinghe FYI.

NipunaRanasinghe: @ThisaruGuruge Yes I agree. Otherwise we'll have keep doing this extra step in all the future connectors as well. +1 to fix this behaviour from the action side itself.

keizer619: Yes we need to fix it there. Once we fix that we should be able to remove these everwhere

PR #15, opened 2024-06-11
Add support for loading static analyzer plugins

Xenowa:
## Purpose

Add functionalities to:

- Load external analyzer compiler plugins defined in s `Scan.toml` file and perform additional analysis.

```toml
[[analyzer]]
org = "exampleOrg"
name = "example_module_static_code_analyzer"

[[analyzer]]
org = "ballerina"
name = "example_module_static_code_analyzer"
version = "0.1.0"

[[analyzer]]
org = "ballerinax"
name = "example_module_static_code_analyzer"
version = "0.1.0"
repository = "local"
```

- print all static code analysis rules to the console.

```bash
        RuleID                                           | Rule Severity | Rule Description 
        ------------------------------------------------------------------------------------
        ballerina:1                                      | CODE_SMELL    | Avoid checkpanic 
        ballerina/example_module_static_code_analyzer:1  | CODE_SMELL    | rule 1           
        ballerina/example_module_static_code_analyzer:2  | BUG           | rule 2           
        ballerina/example_module_static_code_analyzer:3  | VULNERABILITY | rule 3           
        ballerinax/example_module_static_code_analyzer:1 | CODE_SMELL    | rule 1           
        ballerinax/example_module_static_code_analyzer:2 | BUG           | rule 2           
        ballerinax/example_module_static_code_analyzer:3 | VULNERABILITY | rule 3           
        exampleOrg/example_module_static_code_analyzer:1 | CODE_SMELL    | rule 1           
        exampleOrg/example_module_static_code_analyzer:2 | BUG           | rule 2           
        exampleOrg/example_module_static_code_analyzer:3 | VULNERABILITY | rule 3 
```


scan-command/src/main/java/io/ballerina/scan/internal/ProjectAnalyzer.java
```diff
@@ -69,51 +97,119 @@ private Consumer<DocumentId> analyzeDocument(Module module, ScannerContextImpl s
...
+                    URL jarUrl = Path.of(jarPath).toUri().toURL();
+                    jarUrls.add(jarUrl);
+                } catch (MalformedURLException ex) {
+                    throw new RuntimeException(ex);
+                }
+            });
+
+            URLClassLoader ucl = AccessController.doPrivileged((PrivilegedAction<URLClassLoader>) () ->
+                    new URLClassLoader(jarUrls.toArray(new URL[0]), this.getClass().getClassLoader())
+            );
```
heshanpadmasiri, line 160: Do we really need to do this as a privileged action?. Constructor of `URLClassLoader` calls the security manager, thus it may be necessary. Problem is whole `AccessController` class is now deprecated for removal ([ref](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/security/AccessController.html#doPrivileged(java.security.PrivilegedAction))), along with `SecurityManager`. Therefore at least put a comment saying why it is needed

scan-command/src/main/java/io/ballerina/scan/internal/ProjectAnalyzer.java
```diff
@@ -69,51 +97,119 @@ private Consumer<DocumentId> analyzeDocument(Module module, ScannerContextImpl s
...
+            try (BufferedReader reader = new BufferedReader(
+                    new InputStreamReader(resourceAsStream,
+                            StandardCharsets.UTF_8))) {
+                String line;
+                while ((line = reader.readLine()) != null) {
+                    stringBuilder.append(line);
+                }
+            } catch (IOException e) {
+                throw new RuntimeException(e);
+            }
```
heshanpadmasiri, line 176: You can make this more concise by doing a [mutable reduction](https://docs.oracle.com/javase/8/docs/api/java/util/stream/Collector.html), instead of explicitly using a `StringBuilder` and converting to String. Do note this is using the stream API, other than that this should perform more or less the same
```suggestion
            String res;
            try (BufferedReader reader = new BufferedReader(
                    new InputStreamReader(resourceAsStream,
                            StandardCharsets.UTF_8))) {
                res = reader.lines().collect(Collectors.joining());
            } catch (IOException ex) {
                throw new UncheckedIOException(ex);
            }
```

scan-command/src/main/java/io/ballerina/scan/internal/ProjectAnalyzer.java
```diff
@@ -69,51 +97,119 @@ private Consumer<DocumentId> analyzeDocument(Module module, ScannerContextImpl s
...
+                    case "CODE_SMELL" -> RuleKind.CODE_SMELL;
+                    default -> null;
+                };
+
+                String description = ruleObject.get("description").getAsString();
+                if (ruleKind != null) {
+                    Rule inMemoryRule = RuleFactory.createRule(numericId, description, ruleKind,
+                            org.value(), name.value());
+                    externalRules.add(inMemoryRule);
+                }
```
heshanpadmasiri, line 197: Given you use `numericId` and `description` only if `ruleKind` is not null I think you can restrcuture things to be more concise  by moving them closer to where you are actually using them
```suggestion
                                RuleKind ruleKind = switch (ruleObject.get("ruleKind").getAsString()) {
                    case "BUG" -> RuleKind.BUG;
                    case "VULNERABILITY" -> RuleKind.VULNERABILITY;
                    case "CODE_SMELL" -> RuleKind.CODE_SMELL;
                    default -> null;
                };

                if (ruleKind != null) {
                    int numericId = ruleObject.get("id").getAsInt();
                    String description = ruleObject.get("description").getAsString();
                    externalRules.add(RuleFactory.createRule(numericId, description, ruleKind,
                            org.value(), name.value()));
                }

```

scan-command/src/main/java/io/ballerina/scan/internal/ProjectAnalyzer.java
```diff
@@ -133,4 +229,29 @@ private void extractAnalyzerImportsAndDependencies(ScanTomlFile.Analyzer analyze
     private void buildStringWithNewLine(StringBuilder stringBuilder, String content) {
         stringBuilder.append(content).append(System.lineSeparator());
     }
+
+    List<Issue> runExternalAnalyzers(Project project, Map<String, List<Rule>> externalAnalyzers) {
+        List<ScannerContext> scannerContextList = new ArrayList<>();
```
heshanpadmasiri, line 234: I think you can initialize this with initial capacity as well since you add one for each in externalAnalyzers map

scan-command/src/main/java/io/ballerina/scan/internal/ProjectAnalyzer.java
```diff
@@ -133,4 +229,29 @@ private void extractAnalyzerImportsAndDependencies(ScanTomlFile.Analyzer analyze
...
+        for (Map.Entry<String, List<Rule>> externalAnalyzer : externalAnalyzers.entrySet()) {
+            ScannerContextImpl scannerContext = new ScannerContextImpl(externalAnalyzer.getValue());
+            scannerContextList.add(scannerContext);
+            Map<String, Object> pluginProperties = new HashMap<>();
+            pluginProperties.put("ScannerContext", scannerContext);
+            project.projectEnvironmentContext().getService(CompilerPluginCache.class)
+                    .putData(externalAnalyzer.getKey(), pluginProperties);
```
heshanpadmasiri, line 241: Are we populating this map for some side effect? May be put a comment if so

scan-command/src/main/java/io/ballerina/scan/utils/ScanUtils.java
```diff
@@ -550,4 +551,38 @@ private static Optional<String> loadRemoteJAR(Path targetDir, String fileName, U
...
+        String format = "\t%-" + maxRuleIDLength + "s | %-" + maxSeverityLength + "s | %-" + maxDescriptionLength
+                + "s%n";
+
+        outputStream.printf(format, ruleIDColumn, ruleSeverityColumn, ruleDescriptionColumn);
+        outputStream.printf("\t" + "-".repeat(maxRuleIDLength + 1) + "--" +
+                "-".repeat(maxSeverityLength + 1) + "--" + "-".repeat(maxDescriptionLength + 1) + "%n");
+
+        for (Rule rule : rules) {
+            String formattedLine = String.format(format, rule.id(), rule.kind().toString(), rule.description());
+            outputStream.println(formattedLine.replaceAll("\\s+$", ""));
```
heshanpadmasiri, line 585: Are you trying to remove trailing white spaces here? If so I think it is more clear to just use `formattedLine.stripTrailing()`

scan-command/src/main/java/io/ballerina/scan/internal/ProjectAnalyzer.java
```diff
@@ -18,25 +18,52 @@
...
 import io.ballerina.projects.ModuleId;
+import io.ballerina.projects.Package;
+import io.ballerina.projects.PackageDependencyScope;
+import io.ballerina.projects.PackageManifest;
+import io.ballerina.projects.PackageName;
+import io.ballerina.projects.PackageOrg;
+import io.ballerina.projects.PackageResolution;
 import io.ballerina.projects.Project;
+import io.ballerina.projects.ResolvedPackageDependency;
+import io.ballerina.projects.internal.model.CompilerPluginDescriptor;
```
MaryamZi, line 38: Are these internal classes meant to be accessed outside lang? @gayaldassanayake

scan-command/src/main/java/io/ballerina/scan/internal/ProjectAnalyzer.java
```diff
@@ -69,51 +97,119 @@ private Consumer<DocumentId> analyzeDocument(Module module, ScannerContextImpl s
...
         List<String> defaultModuleFiles = defaultModule.documentIds()
                 .stream()
                 .map(documentId -> defaultModule.document(documentId).name())
                 .toList();
-        String documentName;
         do {
-            documentName = String.format("%s-%s.bal", IMPORT_GENERATOR_FILE, UUID.randomUUID());
-        } while (defaultModuleFiles.contains(documentName));
+            generatedDocumentName = String.format("%s-%s.bal", IMPORT_GENERATOR_FILE, UUID.randomUUID());
+        } while (defaultModuleFiles.contains(generatedDocumentName));
```
MaryamZi, line 115: Can't we do this in the constructor?

scan-command/src/main/java/io/ballerina/scan/internal/ProjectAnalyzer.java
```diff
@@ -69,51 +97,119 @@ private Consumer<DocumentId> analyzeDocument(Module module, ScannerContextImpl s
...
+            }
+
+            String fqn = pluginDesc.plugin().getClassName();
+            List<Rule> externalRules = new ArrayList<>();
+            Gson gson = new Gson();
+            JsonArray ruleArray = gson.fromJson(stringBuilder.toString(), JsonArray.class);
+            ruleArray.forEach(rule -> {
+                JsonObject ruleObject = rule.getAsJsonObject();
+                int numericId = ruleObject.get("id").getAsInt();
+                RuleKind ruleKind = switch (ruleObject.get("ruleKind").getAsString()) {
```
MaryamZi, line 185: Introduce constants for values like `"ruleKind"`.

scan-command/src/main/java/io/ballerina/scan/utils/ScanUtils.java
```diff
@@ -550,4 +551,38 @@ private static Optional<String> loadRemoteJAR(Path targetDir, String fileName, U
...
+    /**
+     * Prints the {@link List<Rule>} of static analysis rules to the console.
+     *
+     * @param rules        the list of static analysis rules
+     * @param outputStream the print stream
+     */
+    public static void printRulesToConsole(List<Rule> rules, PrintStream outputStream) {
+        String ruleIDColumn = "RuleID";
+        String ruleSeverityColumn = "Rule Severity";
+        String ruleDescriptionColumn = "Rule Description";
```
MaryamZi, line 564: Can be constants?

scan-command/src/test/resources/command-outputs/print-rules-to-console.txt
```diff
@@ -0,0 +1,13 @@
...
+	ballerina:1                                      | CODE_SMELL    | Avoid checkpanic
+	ballerina/example_module_static_code_analyzer:1  | CODE_SMELL    | rule 1
+	ballerina/example_module_static_code_analyzer:2  | BUG           | rule 2
+	ballerina/example_module_static_code_analyzer:3  | VULNERABILITY | rule 3
+	exampleOrg/example_module_static_code_analyzer:1 | CODE_SMELL    | rule 1
+	exampleOrg/example_module_static_code_analyzer:2 | BUG           | rule 2
+	exampleOrg/example_module_static_code_analyzer:3 | VULNERABILITY | rule 3
+	ballerinax/example_module_static_code_analyzer:1 | CODE_SMELL    | rule 1
+	ballerinax/example_module_static_code_analyzer:2 | BUG           | rule 2
+	ballerinax/example_module_static_code_analyzer:3 | VULNERABILITY | rule 3
```
MaryamZi, line 13: How is the package order decided? IMO, we should prioritize by org - ballerina, ballerinax, wso2, then other orgs alphabetically sorted or completely sort alphabetically.

I prefer the former.

test-compiler-plugins/ballerina-plugin/example-module-static-code-analyzer/CompilerPlugin.toml
```diff
@@ -0,0 +1,5 @@
+[plugin] # Have to check if below path is needed
+class = "io.ballerina.examplestaticcodeanalyzer.CustomStaticCodeAnalyzer"
+
+[[dependency]]
+path = "../build/libs/ballerina-plugin-0.1.0.jar"
```
MaryamZi, line 5:
```suggestion
path = "../build/libs/ballerina-plugin-0.1.0.jar"

```

test-compiler-plugins/ballerina-plugin/example-module-static-code-analyzer/main.bal
```diff
@@ -0,0 +1,5 @@
+import ballerina/io;
```
MaryamZi, line 1: Didn't we decide not to import library modules?

test-compiler-plugins/ballerina-plugin/src/main/java/io/ballerina/examplestaticcodeanalyzer/CustomAnalysisTask.java
```diff
@@ -0,0 +1,44 @@
...
+
+import io.ballerina.compiler.syntax.tree.ModulePartNode;
+import io.ballerina.projects.Document;
+import io.ballerina.projects.Module;
+import io.ballerina.projects.plugins.AnalysisTask;
+import io.ballerina.projects.plugins.SyntaxNodeAnalysisContext;
+import io.ballerina.scan.Reporter;
+import io.ballerina.scan.ScannerContext;
+
+public class CustomAnalysisTask implements AnalysisTask<SyntaxNodeAnalysisContext> {
```
MaryamZi, line 29: Add simple docs + version.

test-compiler-plugins/ballerina-plugin/src/main/resources/rules.json
```diff
@@ -0,0 +1,17 @@
...
+    "id": 2,
+    "ruleKind": "BUG",
+    "description": "rule 2"
+  },
+  {
+    "id": 3,
+    "ruleKind": "VULNERABILITY",
+    "description": "rule 3"
+  }
+]
```
MaryamZi, line 17:
```suggestion
]

```

Fix wherever applicable.

scan-command/src/test/resources/command-outputs/print-rules-to-console.txt
Xenowa: This is based on the following order, where the core rules in the scan tool are displayed first, then for the compiler plugin rules, the order is decided by the project API package resolution.

This is based on the following order:
- Initially, the core rules defined in the scan tool are shown.
- Next, the rules obtained from the compiler plugins are shown in the order determined through the package resolution.

scan-command/src/main/java/io/ballerina/scan/internal/ProjectAnalyzer.java
Xenowa: Had this block previously to remove the spot bugs warning. will replace it with the static method `URLClassLoader.newInstance()` to obtain the class loader instead.

scan-command/src/test/resources/command-outputs/print-rules-to-console.txt
MaryamZi: I would use the order I suggested. Compiler plugin order doesn't mean anything to the user, does it?

scan-command/src/main/java/io/ballerina/scan/internal/ProjectAnalyzer.java
Xenowa: Yes, the static analyzer compiler plugins retrieve the context passed from this map via the `CompilerPluginContext` and perform additional analysis when they get engaged during project compilation

scan-command/src/main/java/io/ballerina/scan/internal/ProjectAnalyzer.java
```diff
@@ -69,68 +116,167 @@ private Consumer<DocumentId> analyzeDocument(Module module, ScannerContextImpl s
         };
     }
 
-    List<Issue> runExternalAnalyzers(Project project) {
+    Optional<Map<String, List<Rule>>> getExternalAnalyzers(PrintStream outputStream) {
```
heshanpadmasiri, line 119: Ideally you should not return an optional collection. Instead you should just return just an empty collection. Otherwise receiver has to handle case where value is empty and collection is empty as two distinct scenarios which could be confusing. If you really want to signal 3 states collection being empty, collection having values and "empty" then it is better to return a record with a boolean flag saying what "empty" means in addition to the collection

scan-command/src/main/java/io/ballerina/scan/internal/ProjectAnalyzer.java
gayaldassanayake: Accessing the `CompilerPluginDescriptor` is fine. However, it is true that having them in a module named `internal` is bit misleading.

scan-command/src/main/java/io/ballerina/scan/internal/ProjectAnalyzer.java
```diff
@@ -69,68 +116,167 @@ private Consumer<DocumentId> analyzeDocument(Module module, ScannerContextImpl s
...
+                            throw new RuntimeException(ex);
+                        }
+            }).toList();
 
-        List<Issue> externalIssues = new ArrayList<>(scannerContextList.size());
-        for (ScannerContext scannerContext : scannerContextList) {
-            ReporterImpl reporter = (ReporterImpl) scannerContext.getReporter();
-            externalIssues.addAll(reporter.getIssues());
+            URLClassLoader ucl = URLClassLoader.newInstance(jarUrls.toArray(new URL[0]),
+                    this.getClass().getClassLoader());
```
heshanpadmasiri, line 163: Instead of creating a empty array use `new` method reference. This way `toArray` will create the array of the size required, where as passing `new URL[0]` will cause you to create an empty array and `toArray` to create another array to hold the values
```suggestion
            URLClassLoader ucl = URLClassLoader.newInstance(jarUrls.toArray(URL[]::new),
                    this.getClass().getClassLoader());
```

scan-command/src/main/java/io/ballerina/scan/utils/ScanUtils.java
```diff
@@ -550,4 +555,91 @@ private static Optional<String> loadRemoteJAR(Path targetDir, String fileName, U
...
+     * </ol>
+     *
+     * @param rules The list of rules to be sorted.
+     */
+    private static void sortRules(List<Rule> rules) {
+        Comparator<Rule> ruleComparator = (r1, r2) -> {
+            String id1 = r1.id().split(":")[0];
+            String id2 = r2.id().split(":")[0];
+
+            if (id1.equals("ballerina") && !id2.equals("ballerina")) {
```
gayaldassanayake, line 613: Why do we have both "ballerina" and "ballerina/" here? Can't we strip `/` when creating id1, id2 and remove the "ballerina/" part?

scan-command/src/main/java/io/ballerina/scan/utils/ScanUtils.java
```diff
@@ -550,4 +555,91 @@ private static Optional<String> loadRemoteJAR(Path targetDir, String fileName, U
...
+     *  <li> `ballerinax/moduleName` rules next </li>
+     *  <li> `wso2/moduleName` rules next </li>
+     *  <li> All remaining rules sorted alphabetically </li>
+     * </ol>
+     *
+     * @param rules The list of rules to be sorted.
+     */
+    private static void sortRules(List<Rule> rules) {
+        Comparator<Rule> ruleComparator = (r1, r2) -> {
+            String id1 = r1.id().split(":")[0];
```
heshanpadmasiri, line 610: I feel like this can be done more cleanly. What you need is a list of prefixes sorted in the priority you need. First check if the prefix is in the list and if so compare by index. Else compare by string values. Otherwise if we want to make changes to the sort order, like adding new stuff, this would be very confusing. For example (change as needed),

```
public int compare(String s1, String s2) {
        // Define the priorities for the prefixes
        List<String> priorities = List.of("ballerina", "ballerina/", "ballerinax/", "wso2/");

        // Determine the priority of each string
        int priority1 = getPriority(s1, priorities);
        int priority2 = getPriority(s2, priorities);

        // Compare based on priority
        if (priority1 != priority2) {
            return Integer.compare(priority1, priority2);
        }

        // If both strings have the same priority, sort alphabetically
        return s1.compareTo(s2);
    }

    private int getPriority(String s, List<String> priorities) {
        return priorities.stream()
            .filter(s::startsWith)
            .findFirst()
            .map(priorities::indexOf)
            .orElse(priorities.size());
    }
```

heshanpadmasiri (commented):
Minor nits otherwise LGTM

scan-command/src/main/java/io/ballerina/scan/utils/ScanUtils.java
SasinduDilshara, line 613: Shall we make this as constants

test-compiler-plugins/exampleOrg-plugin/example-module-static-code-analyzer/main.bal
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) 2024, WSO2 LLC. (https://www.wso2.com).
```
SasinduDilshara, line 1: Is this file empty?

scan-command/src/main/java/io/ballerina/scan/internal/ProjectAnalyzer.java
```diff
@@ -69,68 +116,167 @@ private Consumer<DocumentId> analyzeDocument(Module module, ScannerContextImpl s
...
+        DocumentConfig documentConfig = DocumentConfig.from(documentId, newImports.toString(), generatedDocumentName);
         defaultModule.modify().addDocument(documentConfig).apply();
-
         project.currentPackage().ballerinaToml().ifPresent(ballerinaToml -> {
             String tomlFileContent = ballerinaToml.tomlDocument().textDocument().toString();
             ballerinaToml.modify().withContent(tomlFileContent + tomlDependencies).apply();
         });
 
-        List<ScannerContext> scannerContextList = new ArrayList<>();
+        PackageResolution packageResolution = project.currentPackage().getResolution();
```
prakanth97, line 138: This variable seems redundant.

prakanth97: Shall we extract `project.currentPackage()` to a variable

scan-command/src/main/java/io/ballerina/scan/internal/ProjectAnalyzer.java
```diff
@@ -69,68 +116,167 @@ private Consumer<DocumentId> analyzeDocument(Module module, ScannerContextImpl s
...
+        ResolvedPackageDependency rootPkgNode = new ResolvedPackageDependency(project.currentPackage(),
+                PackageDependencyScope.DEFAULT);
+        List<Package> directDependencies = packageResolution.dependencyGraph().getDirectDependencies(rootPkgNode)
+                .stream().map(ResolvedPackageDependency::packageInstance).toList();
+        Map<String, List<Rule>> externalAnalyzers = new HashMap<>();
+        for (Package pkgDependency : directDependencies) {
+            PackageManifest pkgManifest = pkgDependency.manifest();
+            String org = pkgManifest.org().value();
+            String name = pkgManifest.name().value();
```
prakanth97, line 147: These variables are not required since it is used in single place.

scan-command/src/main/java/io/ballerina/scan/internal/ProjectAnalyzer.java
```diff
@@ -69,68 +116,167 @@ private Consumer<DocumentId> analyzeDocument(Module module, ScannerContextImpl s
...
+
+            String output;
+            try (BufferedReader reader = new BufferedReader(new InputStreamReader(resourceAsStream,
+                    StandardCharsets.UTF_8))) {
+                output = reader.lines().collect(Collectors.joining());
+            } catch (IOException ex) {
+                throw new RuntimeException(ex);
+            }
+
+            String fqn = pluginDesc.plugin().getClassName();
```
prakanth97, line 177: This variable seems redundant, can you check this and all places and remove them. 

Also, create a variable closer to its usage, otherwise it will stay in the memory for unnecessary execution.

scan-command/src/main/java/io/ballerina/scan/internal/ProjectAnalyzer.java
```diff
@@ -69,68 +116,167 @@ private Consumer<DocumentId> analyzeDocument(Module module, ScannerContextImpl s
...
+            List<Rule> externalRules = new ArrayList<>();
+            Gson gson = new GsonBuilder().setPrettyPrinting().create();
+            JsonElement element = gson.fromJson(output, JsonElement.class);
+            if (!element.isJsonArray()) {
+                outputStream.println("Invalid JSON format for " + RULES_FILE + " in plugin '" + pluginName + "'.");
+                outputStream.println("Expected an array, but found:");
+                outputStream.println(gson.toJson(element));
+                return Optional.empty();
```
prakanth97, line 185: 1. Shall we move this into a separate method and pass these messages as parameters? this method can be used for the L192 - L195 as well.
2. Make these String a constant

scan-command/src/main/java/io/ballerina/scan/internal/ProjectAnalyzer.java
```diff
@@ -69,68 +116,167 @@ private Consumer<DocumentId> analyzeDocument(Module module, ScannerContextImpl s
...
+                    return Optional.empty();
+                }
+
+                String kind = ruleObject.get(RULE_KIND).getAsString();
+                RuleKind ruleKind = switch (kind) {
+                    case BUG -> RuleKind.BUG;
+                    case VULNERABILITY -> RuleKind.VULNERABILITY;
+                    case CODE_SMELL -> RuleKind.CODE_SMELL;
+                    default -> null;
+                };
```
prakanth97, line 204: Extract this into a separate method

scan-command/src/main/java/io/ballerina/scan/internal/ProjectAnalyzer.java
```diff
@@ -69,68 +116,167 @@ private Consumer<DocumentId> analyzeDocument(Module module, ScannerContextImpl s
...
+                    this.getClass().getClassLoader());
+            InputStream resourceAsStream = ucl.getResourceAsStream(RULES_FILE);
+            if (resourceAsStream == null) {
+                continue;
+            }
+
+            String output;
+            try (BufferedReader reader = new BufferedReader(new InputStreamReader(resourceAsStream,
+                    StandardCharsets.UTF_8))) {
+                output = reader.lines().collect(Collectors.joining());
```
LakshanWeerasinghe, line 172: need to close the bufferedReader

scan-command/src/main/java/io/ballerina/scan/internal/ProjectAnalyzer.java
```diff
@@ -69,68 +116,167 @@ private Consumer<DocumentId> analyzeDocument(Module module, ScannerContextImpl s
...
+                };
+                if (ruleKind == null) {
+                    outputStream.println("Invalid RuleKind for rule object in plugin '" + pluginName + "'.");
+                    outputStream.println("RuleKind should be one of " + Arrays.toString(RuleKind.values()) +
+                            ", but found: " + kind);
+                    return Optional.empty();
+                }
+                int numericId = ruleObject.get(RULE_ID).getAsInt();
+                String description = ruleObject.get(RULE_DESCRIPTION).getAsString();
+                Rule inMemoryRule = RuleFactory.createRule(numericId, description, ruleKind, org, name);
```
prakanth97, line 213: Remove this variables and inline them

scan-command/src/main/java/io/ballerina/scan/utils/ScanUtils.java
Xenowa: There are two types of rules used by the scan tool,

- Core rules
- External rules

The fully qualified rule names of core rules follows the format `ballerina:1` while external rules follow the format `org/name:1`

So for example if the ballerina io module introduces static analysis rules via a compiler plugin, there fully qualified rule identifier would look like `ballerina/io:1`, and  if there are rules introduced by the ballerina http module it would be as `ballerina/http:1`

Also, we should give priority to sorting in the order:
1. Core rules (rules built into the scan tool)
2. ballerina org compiler plugin rules
3. ballerinax org compiler plugin rules
4. wso2 org compiler plugin rules
5. external org compiler plugin rules.

So if we also have rules introduced by an external compiler plugin where the org is arc and the name is http, the printed output should be ordered in the following manner:
```bash
ballerina:1
ballerina/http:1
ballerina/io:1
arc/http:1
```

if we remove the `ballerina/` part, then This order cannot be preserved.

test-compiler-plugins/exampleOrg-plugin/example-module-static-code-analyzer/main.bal
Xenowa: Yes,

When creating compiler plugins and packaging them using the `bal pack` command, the process [checks if a Ballerina file is present in the current package](https://github.com/ballerina-platform/ballerina-lang/blob/master/cli/ballerina-cli/src/main/java/io/ballerina/cli/cmd/PackCommand.java#L263). To ensure the packaging runs without errors, this file is included.

scan-command/src/main/java/io/ballerina/scan/internal/ProjectAnalyzer.java
Xenowa: The `org` and `name` variables are used again during the [creation of the in-memory rules](https://github.com/ballerina-platform/static-code-analysis-tool/pull/15/files/8daa975515a8e8826a7c2d48749dce05dac5eb3d#diff-8fbdfbacc9f46bf89493e82b5eb294ea34cf91175dd952080f3bafdc4139d31eR213)

scan-command/src/main/java/io/ballerina/scan/utils/ScanUtils.java
```diff
@@ -550,4 +555,91 @@ private static Optional<String> loadRemoteJAR(Path targetDir, String fileName, U
...
+     *  <li> `ballerina` rules first </li>
+     *  <li> `ballerina/moduleName` rules next </li>
+     *  <li> `ballerinax/moduleName` rules next </li>
+     *  <li> `wso2/moduleName` rules next </li>
+     *  <li> All remaining rules sorted alphabetically </li>
+     * </ol>
+     *
+     * @param rules The list of rules to be sorted.
+     */
+    private static void sortRules(List<Rule> rules) {
```
LakshanWeerasinghe, line 608: Shall we add unit test to test this util function?

scan-command/src/main/java/io/ballerina/scan/internal/ProjectAnalyzer.java
Xenowa: When engaging compiler plugins through the project API, the following steps are followed:

1. Generate an in-memory Ballerina file in the default module with the compiler plugins generated as imports.
2. Modify the in-memory `Ballerina.toml` file by generating the local compiler plugin toml dependencies.
3. Engage the generated compiler plugins through package compilation.

However, the project API allows doing one in-memory modification step at a time, and the subsequent steps to the modified project are applied only if the `currentPackage` is loaded again from the `project` object.

image


Therefore we cannot extract the `currentPackage` to a re-usable variable.

scan-command/src/main/java/io/ballerina/scan/internal/ProjectAnalyzer.java
```diff
@@ -18,40 +18,87 @@
...
+        this.project = project;
         this.scanTomlFile = scanTomlFile;
+        Module defaultModule = project.currentPackage().getDefaultModule();
+        List<String> defaultModuleFiles = defaultModule.documentIds()
+                .stream()
+                .map(documentId -> defaultModule.document(documentId).name())
+                .toList();
+        do {
+            generatedDocumentName = String.format("%s-%s.bal", IMPORT_GENERATOR_FILE, UUID.randomUUID());
+        } while (defaultModuleFiles.contains(generatedDocumentName));
```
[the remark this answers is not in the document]
Xenowa: There  are two reasons for this:

1. To make sure to filter out any analysis issues generated on the import-generated Ballerina file.
2. If the file name of the import generation file is same as an actual file in the default module, it will generate static analysis issues with incorrect line numbers.

scan-command/src/main/java/io/ballerina/scan/internal/ProjectAnalyzer.java
```diff
@@ -18,40 +18,87 @@
...
 class ProjectAnalyzer {
+    private final Project project;
     private final ScanTomlFile scanTomlFile;
+    private String generatedDocumentName;
 
-    ProjectAnalyzer(ScanTomlFile scanTomlFile) {
+    ProjectAnalyzer(Project project, ScanTomlFile scanTomlFile) {
+        this.project = project;
         this.scanTomlFile = scanTomlFile;
+        Module defaultModule = project.currentPackage().getDefaultModule();
```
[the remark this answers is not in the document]
Xenowa: Explained in https://github.com/ballerina-platform/static-code-analysis-tool/pull/15#discussion_r1639731937

scan-command/src/main/java/io/ballerina/scan/internal/ProjectAnalyzer.java
Xenowa: This change will not be done in preference to https://github.com/ballerina-platform/static-code-analysis-tool/pull/15#discussion_r1637626715

gayaldassanayake (approved):

heshanpadmasiri (approved):
LGTM

scan-command/src/main/resources/scan.properties
```diff
@@ -0,0 +1,74 @@
...
+#
+
+# ------------------------
+# Scan Tool error messages
+# ------------------------
+error.invalid.number.of.arguments=\
+  Invalid number of arguments, expected one argument received {0}
+
+error.empty.package=\
+  ballerina: Package is empty. Please add at least one .bal file.
```
MaryamZi, line 26: Shall we avoid the `ballerina:`? Since this isn't a core error.

scan-command/src/main/java/io/ballerina/scan/utils/DiagnosticLog.java
```diff
@@ -0,0 +1,47 @@
...
+import java.text.MessageFormat;
+import java.util.Locale;
+import java.util.ResourceBundle;
+
+/**
+ * Diagnostic log for scan tool.
+ *
+ * @since 0.1.0
+ */
+public class DiagnosticLog {
```
MaryamZi, line 30: Do you have tests for these errors?

scan-command/src/main/resources/scan.properties
```diff
@@ -0,0 +1,74 @@
...
+
+error.missing.config.field=\
+  Failed to load scan tool configurations: {0} is missing
```
MaryamZi, line 29:
```suggestion
  Failed to load scan tool configurations: cannot find {0}
```

What happens if multiple are unavailable?

scan-command/src/main/resources/scan.properties
```diff
@@ -0,0 +1,74 @@
...
+
+error.loading.remote.config.file=\
+  Failed to load configuration file with exception: {0}
```
MaryamZi, line 32: Won't these errors be shown to the user? Let's avoid terms like exceptions, which aren't used in the Ballerina world.

scan-command/src/main/resources/scan.properties
```diff
@@ -0,0 +1,74 @@
...
+
+error.reading.config.file=\
+  Failed to read the configuration file with exception: {0}
```
MaryamZi, line 35:
```suggestion
  Failed to read the configuration file: {0}
```

scan-command/src/main/resources/scan.properties
```diff
@@ -0,0 +1,74 @@
...
+  Failed to download remote JAR file: {0}
+
+error.invalid.json.format=\
+  Invalid JSON format for {0} in plugin '{1}'.\
+  Expected an array, but found:\
+  {2}
+
+error.invalid.json.format.rule=\
+  Invalid JSON format for rule object in plugin '{0}'.\
+  Rule object should contain numeric id, ruleKind and description, but found:\
```
MaryamZi, line 50: Use the exact key everywhere and emphasize. E.g., 'ruleKind'

scan-command/src/main/resources/scan.properties
```diff
@@ -0,0 +1,74 @@
...
+  {1}
+
+error.invalid.json.format.rule.kind=\
+  Invalid RuleKind for rule object in plugin '{0}'.\
+  RuleKind should be one of {1}, but found:\
```
MaryamZi, line 55: ruleKind vs RuleKind is not consistent. Please recheck all diagnostics for consistency.

scan-command/src/main/resources/scan.properties
```diff
@@ -0,0 +1,74 @@
...
+  Failed to copy the file with exception: {0}
+
+error.rule.not.found=\
+  Rule not found: Invalid rule numeric identifier '{0}'.
+
+# --------------------------
+# Scan Tool warning messages
+# --------------------------
+warning.report.not.supported=\
+  Generating reports is not supported with single bal files. Ignoring the flag and continuing the scans...
```
MaryamZi, line 71: We can omit ... at the end.

scan-command/src/main/resources/scan.properties
```diff
@@ -0,0 +1,74 @@
...
+  {2}
+
+error.failed.to.read.file=\
+  Failed to read the file with exception: {0}
+
+error.failed.to.copy.file=\
+  Failed to copy the file with exception: {0}
+
+error.rule.not.found=\
+  Rule not found: Invalid rule numeric identifier '{0}'.
```
MaryamZi, line 65: Period at the end not consistent.

scan-command/src/main/java/io/ballerina/scan/utils/DiagnosticLog.java
Xenowa: Yes, they have been added in the [DiagnosticLogTest class](https://github.com/ballerina-platform/static-code-analysis-tool/pull/15/files#diff-096e1e7f096c7da225cdf5c19778594d9e825bc369ac841798b8e41b793d7cc9)

scan-command/src/main/resources/scan.properties
Xenowa: The `config` field is used for pointing to the `Scan.toml` configuration file from the `Ballerina.toml` file.

```toml
# Ballerina.toml
[scan]
config = "path/to/Scan.toml"
```

So this error was made to be shown if the user does not define the path to the configuration file as follows:

```toml
# Ballerina.toml
[scan]
# Missing config field
```

Should we consider that the  user will point to multiple configuration files?

MaryamZi: Can you share what the diagnostic will look like?

scan-command/src/main/java/io/ballerina/scan/utils/DiagnosticLog.java
MaryamZi: Yeah, but there we don't seem to have tests for all possible diagnostics?

scan-command/src/main/resources/scan.properties
Xenowa: Yes,

If we consider a Ballerina project:
```
📦ballerina-project
 ┗ 📜.devcontainer.json
 ┗ 📜.gitignore
 ┗ 📜Ballerina.toml
 ┗ 📜main.bal
 ┗ 📜Scan.toml
```

With the `Ballerina.toml` defined as:
```toml
[scan]
# configPath is missing
```

The result displayed in the console would be as follows:
```bash
PS C:\Users\Tharana Wanigaratne\Desktop\ballerina-project> bal scan

Running Scans...
Failed to load scan tool configurations: cannot find configPath
```

MaryamZi: So the error message is fixed, right? Why do we need to pass configPath as a param.

Btw, use quotes with configPath. We should either refer directly by name with quotes or have it as a proper term.

Xenowa: Since it's a field to be checked in the toml file, in the Utility method [it's used as a static constant](https://github.com/ballerina-platform/static-code-analysis-tool/blob/16c40747349c5d92f3ddce4f6ba79b1b20ebd32e/scan-command/src/main/java/io/ballerina/scan/utils/Constants.java#L47), This is why it was provided as a param.

Will remove the param and have it there as a fixed message.

scan-command/src/main/java/io/ballerina/scan/internal/ProjectAnalyzer.java
dulajdilshan: Got it!. It means you are changing the current-package separately and you expect that change.

scan-command/src/main/resources/scan.properties
```diff
@@ -0,0 +1,69 @@
...
+error.reading.config.file=\
+  Failed to read the configuration file: {0}
+
+error.loading.remote.platform.file=\
+  Failed to retrieve remote platform file: {0}
+
+error.downloading.remote.jar.file=\
+  Failed to download remote JAR file: {0}
+
+error.invalid.json.format=\
```
dulajdilshan, line 43: In ballerina-lang, we have used all-simple cases for error messages. Shouldn't we follow the same behaviour here too?

scan-command/src/test/java/io/ballerina/scan/internal/ProjectAnalyzerTest.java
```diff
@@ -93,9 +85,21 @@ void testAnalyzingProjectWithCoreAnalyzer() {
 
     @Test(description = "Test analyzing project with external analyzers")
     void testAnalyzingProjectWithExternalAnalyzers() {
-        // TODO: Implement mock external analyzer issue assertions.
-        List<Issue> issues = projectAnalyzer.runExternalAnalyzers(project);
-        Assert.assertEquals(issues.size(), 0);
+        Assert.assertNotNull(projectAnalyzer);
```
dulajdilshan, line 88: In which cases can this be `null ' except the RuntimeException occurred in the initialization? Isn't it itself a check? I mean do we really need this null check here as well?

scan-command/src/test/java/io/ballerina/scan/internal/ScanCmdTest.java
```diff
@@ -250,4 +248,94 @@ void testScanCommandGenerateScanReportMethodWhenIssuePresent() throws IOExceptio
...
+        Assert.assertEquals(rules.get(6).id(), "ballerinax/exampleModule:1");
+        Assert.assertEquals(rules.get(7).id(), "ballerinax/exampleModule:2");
+        Assert.assertEquals(rules.get(8).id(), "ballerinax/exampleModule:3");
+        Assert.assertEquals(rules.get(9).id(), "wso2/exampleModule:1");
+        Assert.assertEquals(rules.get(10).id(), "wso2/exampleModule:2");
+        Assert.assertEquals(rules.get(11).id(), "wso2/exampleModule:3");
+        Assert.assertEquals(rules.get(12).id(), "exampleOrg/exampleModule:1");
+        Assert.assertEquals(rules.get(13).id(), "exampleOrg/exampleModule:2");
+        Assert.assertEquals(rules.get(14).id(), "exampleOrg/exampleModule:3");
+    }
```
dulajdilshan, line 340: Shall we use a data-provider function to parse the arguments?

Check this out:

https://github.com/ballerina-platform/ballerina-lang/blob/master/tests/ballerina-compiler-api-test/src/test/java/io/ballerina/semantic/api/test/statements/SymbolsInCallStatementsTest.java#L60-L86

scan-command/src/test/java/io/ballerina/scan/utils/DiagnosticLogTest.java
```diff
@@ -0,0 +1,108 @@
...
+/**
+ * Diagnostic log tests.
+ *
+ * @since 0.1.0
+ */
+public class DiagnosticLogTest extends BaseTest {
+    private final Path commandOutputs = testResources.resolve("command-outputs");
+
+    @Test(description = "Test error diagnostic message")
+    public void testErrorDiagnostic() throws IOException {
```
dulajdilshan, line 44: Check how we are implementing negative tests in here and refactor the test.

https://github.com/ballerina-platform/ballerina-lang/blob/master/tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/expressions/mappingconstructor/MappingConstructorExprTest.java#L70-L102

scan-command/src/test/java/io/ballerina/scan/internal/ProjectAnalyzerTest.java
Xenowa: This was previously implemented to prevent spot bug errors. Tested after removing the null checks and it builds without issues.

scan-command/src/test/java/io/ballerina/scan/utils/DiagnosticLogTest.java
Xenowa: The `BCompileUtil` is used for generating diagnostic errors after compiling Single Ballerina files. However, most error messages in the scan tool are displayed before compiling the project. Only the `RuleNotFound` error is created after a compilation when external compiler plugins are engaged.

scan-command/src/test/java/io/ballerina/scan/utils/DiagnosticLogTest.java
```diff
@@ -0,0 +1,108 @@
...
+        String expected = "Invalid number of arguments, expected one argument received 0";
+        String result = DiagnosticLog.error(DiagnosticCode.INVALID_NUMBER_OF_ARGUMENTS, 0);
+        Assert.assertEquals(result, expected);
```
MaryamZi, line 47: What I meant was actual tests for these error scenarios.

scan-command/src/test/java/io/ballerina/scan/utils/DiagnosticLogTest.java
dulajdilshan: Yeah. We can't do as in the example I provided. We can use a data provider for this if we are testing the diagnostic logger

scan-command/src/main/java/io/ballerina/scan/internal/ExternalAnalyzerResult.java
```diff
@@ -16,21 +16,19 @@
...
-     * Returns a new instance of the RuleNotFoundException with the specified rule identifier.
-     *
-     * @param ruleId numeric identifier of the static code analysis rule
-     */
-    public RuleNotFoundException(int ruleId) {
-        super(String.format("Rule not found: Invalid rule numeric identifier '%d'.", ruleId));
-    }
-}
+ */
+record ExternalAnalyzerResult(Map<String, List<Rule>> externalAnalyzers, boolean hasAnalyzerPluginIssue) { }
```
MaryamZi, line 34: Isn't hasAnalyzerPluginIssue more like failed?

Taking a step back, isn't this more like a union case? We either get the map of analyzers or an error. Do we really need a record here? Can't you just represent failure with an exception to then handle it in one place and return instead?

scan-command/src/main/java/io/ballerina/scan/internal/ProjectAnalyzer.java
```diff
@@ -18,40 +18,88 @@
...
 
 /**
  * Represents the project analyzer used for analyzing projects.
  *
  * @since 0.1.0
  * */
 class ProjectAnalyzer {
+    private final Project project;
     private final ScanTomlFile scanTomlFile;
+    private String generatedDocumentName;
```
MaryamZi, line 87: Rename this to include something about plugin imports.

scan-command/src/test/java/io/ballerina/scan/internal/ScanCmdTest.java
```diff
@@ -250,4 +248,94 @@ void testScanCommandGenerateScanReportMethodWhenIssuePresent() throws IOExceptio
...
+
+    @Test(description = "test method for printing static code analysis rules to the console")
+    void testPrintRulesToConsole() throws IOException {
+        Path ballerinaProject = testResources.resolve("test-resources")
+                .resolve("bal-project-with-config-file");
+        Project project = BuildProject.load(ballerinaProject);
+        System.setProperty("user.dir", ballerinaProject.toString());
+        ScanTomlFile scanTomlFile = ScanUtils.loadScanTomlConfigurations(project, printStream).orElse(null);
+        System.setProperty("user.dir", userDir);
+        Assert.assertNotNull(scanTomlFile);
```
dulajdilshan, line 260: Shall we check this null just after initializing the `scanTomlFile`

scan-command/src/main/java/io/ballerina/scan/internal/ProjectAnalyzer.java
```diff
@@ -69,68 +117,163 @@ private Consumer<DocumentId> analyzeDocument(Module module, ScannerContextImpl s
...
+
+    List<Issue> runExternalAnalyzers(Map<String, List<Rule>> externalAnalyzers) {
+        List<ScannerContext> scannerContextList = new ArrayList<>(externalAnalyzers.size());
+        for (Map.Entry<String, List<Rule>> externalAnalyzer : externalAnalyzers.entrySet()) {
+            ScannerContextImpl scannerContext = new ScannerContextImpl(externalAnalyzer.getValue());
+            scannerContextList.add(scannerContext);
+
+            // Save the scanner context to plugin cache for the compiler plugin to use during package compilation
+            Map<String, Object> pluginProperties = new HashMap<>();
+            pluginProperties.put("ScannerContext", scannerContext);
```
MaryamZi, line 260: Use a constant.

scan-command/src/main/java/io/ballerina/scan/internal/ProjectAnalyzer.java
```diff
@@ -69,68 +117,163 @@ private Consumer<DocumentId> analyzeDocument(Module module, ScannerContextImpl s
...
+
+        project.currentPackage().getCompilation();
+        List<Issue> externalIssues = new ArrayList<>(scannerContextList.size());
+        for (ScannerContext scannerContext : scannerContextList) {
+            ReporterImpl reporter = (ReporterImpl) scannerContext.getReporter();
+            for (Issue issue : reporter.getIssues()) {
+                if (issue.location().lineRange().fileName().equals(generatedDocumentName)) {
+                    continue;
+                }
+                externalIssues.add(issue);
```
MaryamZi, line 274: Check the inverse and add within the if block instead.

MaryamZi: Since it's a simple addition.

scan-command/src/main/java/io/ballerina/scan/internal/ReporterImpl.java
```diff
@@ -58,7 +59,7 @@ class ReporterImpl implements Reporter {
     public void reportIssue(Document reportedDocument, Location location, int ruleId) {
         Rule rule = rules.get(ruleId);
         if (rule == null) {
-            throw new RuleNotFoundException(ruleId);
+            throw new RuntimeException(DiagnosticLog.error(DiagnosticCode.RULE_NOT_FOUND, ruleId));
```
MaryamZi, line 62: Do we need a custom exception for these?

scan-command/src/test/java/io/ballerina/scan/internal/ScanCmdTest.java
```diff
@@ -250,4 +248,94 @@ void testScanCommandGenerateScanReportMethodWhenIssuePresent() throws IOExceptio
...
+            validationResultsFilePath = testResources.resolve("command-outputs")
+                    .resolve("print-rules-to-console.txt");
+        } else {
+            validationResultsFilePath = testResources.resolve("command-outputs")
+                    .resolve("ubuntu").resolve("print-rules-to-console.txt");
+        }
+        String expected = Files.readString(validationResultsFilePath, StandardCharsets.UTF_8)
+                .replace(WINDOWS_LINE_SEPARATOR, LINUX_LINE_SEPARATOR);
+        String result = readOutput(true).trim();
+        Assert.assertEquals(result, expected);
```
dulajdilshan, line 278:
```suggestion
        Assert.assertEquals(eadOutput(true).trim(), expected);
```

scan-command/src/main/java/io/ballerina/scan/internal/ExternalAnalyzerResult.java
MaryamZi: That and what I've suggested aren't the same, right? Think of how it is handled, you will get the map only if there was no failures.

scan-command/src/main/java/io/ballerina/scan/internal/ReporterImpl.java
MaryamZi: What I suggested was having something like ScanToolException for all scan tool related runtime exceptions.

scan-command/src/main/java/io/ballerina/scan/internal/ProjectAnalyzer.java
```diff
@@ -69,68 +117,160 @@ private Consumer<DocumentId> analyzeDocument(Module module, ScannerContextImpl s
...
+                continue;
+            }
 
-        // TODO: Implement the logic to pass scanner contexts to compiler plugins and perform external static analysis.
+            CompilerPluginDescriptor pluginDesc = pkgManifest.compilerPluginDescriptor().get();
+            List<URL> jarUrls = pluginDesc.dependencies().stream().map(dependency -> {
+                        try {
+                            return Path.of(dependency.getPath()).toUri().toURL();
+                        } catch (MalformedURLException ex) {
+                            throw new RuntimeException(ex);
```
MaryamZi, line 160: Better to use a custom message so that we know what exactly failed where. Also, we shouldn't be exposing Java exceptions if we can give a better and at least equally descriptive error.

MaryamZi: Can't this be ScanToolException too? We can uniformly handle `ScanToolException`s and consider everything else as unexpected failures then.

scan-command/src/main/java/io/ballerina/scan/internal/ProjectAnalyzer.java
```diff
@@ -69,68 +117,160 @@ private Consumer<DocumentId> analyzeDocument(Module module, ScannerContextImpl s
...
+            }).toList();
 
-        List<Issue> externalIssues = new ArrayList<>(scannerContextList.size());
-        for (ScannerContext scannerContext : scannerContextList) {
-            ReporterImpl reporter = (ReporterImpl) scannerContext.getReporter();
-            externalIssues.addAll(reporter.getIssues());
+            URLClassLoader ucl = URLClassLoader.newInstance(jarUrls.toArray(URL[]::new),
+                    this.getClass().getClassLoader());
+            InputStream resourceAsStream = ucl.getResourceAsStream(RULES_FILE);
+            if (resourceAsStream == null) {
```
MaryamZi, line 167: Is this being null an acceptable possibility?

scan-command/src/main/java/io/ballerina/scan/internal/ProjectAnalyzer.java
```diff
@@ -69,68 +117,160 @@ private Consumer<DocumentId> analyzeDocument(Module module, ScannerContextImpl s
...
+
+            List<Rule> externalRules = new ArrayList<>();
+            Gson gson = new Gson();
+            JsonElement element = gson.fromJson(output, JsonElement.class);
+            if (!element.isJsonArray()) {
+                throw new ScanToolException(DiagnosticLog.error(DiagnosticCode.INVALID_JSON_FORMAT, RULES_FILE,
+                        pluginName, gson.toJson(element)));
+            }
+
+            JsonArray ruleArray = element.getAsJsonArray();
```
MaryamZi, line 187: I would break this up to methods where possible. 

For example `getRuleJsonArray`. This makes the code more readable giving an overview of what happens and we'll have to check the actual logic only if required.

scan-command/src/main/java/io/ballerina/scan/internal/ProjectAnalyzer.java
```diff
@@ -69,68 +117,160 @@ private Consumer<DocumentId> analyzeDocument(Module module, ScannerContextImpl s
...
+                    default -> {
+                        throw new ScanToolException(DiagnosticLog.error(DiagnosticCode.INVALID_JSON_FORMAT_RULE_KIND,
+                                pluginName, Arrays.toString(RuleKind.values()), kind));
+                    }
+                };
+                Rule inMemoryRule = RuleFactory.createRule(ruleObject.get(RULE_ID).getAsInt(),
+                        ruleObject.get(RULE_DESCRIPTION).getAsString(), ruleKind, org, name);
+                externalRules.add(inMemoryRule);
+            }
+            externalAnalyzers.put(pluginDesc.plugin().getClassName(), externalRules);
```
MaryamZi, line 210: I'd actually extract this entire block out to a method.

scan-command/src/main/resources/scan.properties
```diff
@@ -0,0 +1,69 @@
...
+  rule object should contain ''numeric id'', ''ruleKind'', and ''description'' but found: {1}
+
+error.invalid.json.format.rule.kind=\
+  invalid ruleKind for rule object in plugin ''{0}''. ruleKind should be one of {1} but found: ''{2}''
+
+error.failed.to.read.file=\
+  failed to read the file with exception: {0}
+
+error.failed.to.copy.file=\
+  failed to copy the file with exception: {0}
```
MaryamZi, line 57: Let's avoid terms like exceptions.

scan-command/src/main/java/io/ballerina/scan/internal/ProjectAnalyzer.java
Xenowa: Yes, If the `rules.json` file is not included in a compiler plugin, the outcome of `ucl.getResourceAsStream()` becomes `null`

MaryamZi: Yeah, my question is if that is acceptable? In that case shouldn't there be an error (or warning) somewhere? You're just continuing here atm?

Xenowa: > Yeah, my question is if that is acceptable? In that case shouldn't there be an error (or warning) somewhere? You're just continuing here atm?

The suggestion is correct, will include an exception at this point.

scan-command/src/main/resources/scan.properties
```diff
@@ -0,0 +1,78 @@
...
+  failed to copy the scan report: {0}
+
+error.rule.not.found=\
+  rule not found: invalid rule numeric identifier ''{0}''
+
+# --------------------------
+# Scan Tool warning messages
+# --------------------------
+warning.report.not.supported=\
+  generating reports is not supported with single bal files. ignoring the flag and continuing the scans
```
MaryamZi, line 75:
```suggestion
  generating reports is not supported with single bal files, ignoring the flag and continuing the scans
```

scan-command/src/main/resources/scan.properties
```diff
@@ -0,0 +1,78 @@
...
+
+warning.scan.report.not.supported=\
+  generating scan reports is not supported with single bal files. ignoring the flag and continuing the scans
```
MaryamZi, line 78:
```suggestion
  generating scan reports is not supported with single bal files, ignoring the flag and continuing the scans
```

MaryamZi (approved):
Please address suggestions in a separate PR.

PR #30, opened 2024-11-01
Add static rule for non isolated public constructs

SasinduDilshara:
## Purpose

> Rule description - https://docs.google.com/document/d/117SJltpBJU4lb5hzhTZQe2rTOig_5O97tqetRzHeyU8/edit?tab=t.0#heading=h.97cyv03xqaf5


scan-command/src/main/java/io/ballerina/scan/internal/StaticCodeAnalyzer.java
```diff
@@ -100,6 +116,77 @@ public void visit(ImplicitAnonymousFunctionExpressionNode implicitAnonymousFunct
...
+    private boolean getQualifier(List<Qualifier> qualifierList, String qualifierValue) {
+        for (Qualifier qualifier : qualifierList) {
+            if (qualifier.getValue().equals(qualifierValue)) {
+                return true;
+            }
+        }
+        return false;
+    }
+
+    private boolean getQualifier(NodeList<Token> qualifierList, SyntaxKind qualifier) {
```
LakshanWeerasinghe, line 181:
```suggestion
    private boolean hasQualifier(NodeList<Token> qualifierList, SyntaxKind qualifier) {
```

scan-command/src/main/java/io/ballerina/scan/internal/StaticCodeAnalyzer.java
```diff
@@ -100,6 +116,77 @@ public void visit(ImplicitAnonymousFunctionExpressionNode implicitAnonymousFunct
...
+                }
+                member.accept(this);
+            });
+        });
+    }
+
+    private void checkNonIsolatedPublicMethods(FunctionDefinitionNode member,
+                                               ClassDefinitionNode classDefinitionNode) {
+        semanticModel.symbol(classDefinitionNode).ifPresent(symbol -> {
+            if (symbol instanceof ClassSymbol classSymbol) {
```
LakshanWeerasinghe, line 151: Shall we check the object symbol?

scan-command/src/main/java/io/ballerina/scan/internal/StaticCodeAnalyzer.java
```diff
@@ -75,6 +87,10 @@ public void visit(CheckExpressionNode checkExpressionNode) {
     @Override
     public void visit(FunctionDefinitionNode functionDefinitionNode) {
         checkUnusedFunctionParameters(functionDefinitionNode.functionSignature());
+        if (!functionDefinitionNode.functionName().text().equals(MAIN_METHOD) &&
```
poorna2152, line 90: Shall we extract `functionDefinitionNode.functionName().text()` into a variable

scan-command/src/main/java/io/ballerina/scan/internal/StaticCodeAnalyzer.java
```diff
@@ -100,6 +116,77 @@ public void visit(ImplicitAnonymousFunctionExpressionNode implicitAnonymousFunct
...
+    public void visit(ClassDefinitionNode classDefinitionNode) {
+        checkNonIsolatedPublicClassDefinitions(classDefinitionNode);
+        checkNonIsolatedPublicClassMembers(classDefinitionNode);
+    }
+
+    private void checkNonIsolatedPublicClassDefinitions(ClassDefinitionNode classDefinitionNode) {
+        semanticModel.symbol(classDefinitionNode).ifPresent(symbol -> {
+            if (symbol instanceof ClassSymbol classSymbol) {
+                List<Qualifier> qualifiers = classSymbol.qualifiers();
+                if (getQualifier(qualifiers, PUBLIC_KEYWORD.stringValue()) &&
```
poorna2152, line 129: Can we do this `.stringValue()` function call inside the `getQualifier` method

scan-command/src/main/java/io/ballerina/scan/internal/StaticCodeAnalyzer.java
```diff
@@ -100,6 +116,77 @@ public void visit(ImplicitAnonymousFunctionExpressionNode implicitAnonymousFunct
...
+                    reportIssue(classDefinitionNode, CoreRule.PUBLIC_NON_ISOLATED_CONSTRUCT);
+                }
+            }
+        });
+    }
+
+    private void checkNonIsolatedPublicClassMembers(ClassDefinitionNode classDefinitionNode) {
+        classDefinitionNode.members().forEach(member -> {
+            semanticModel.symbol(member).ifPresent(symbol -> {
+                if (symbol.kind() == SymbolKind.METHOD) {
```
poorna2152, line 140: Previously this `if (symbol instanceof ClassSymbol classSymbol) {` instanceof check was used to cast the symbol we do the same thing here and be consistent. WDYT?

scan-command/src/main/java/io/ballerina/scan/internal/StaticCodeAnalyzer.java
```diff
@@ -100,6 +116,77 @@ public void visit(ImplicitAnonymousFunctionExpressionNode implicitAnonymousFunct
...
+            if (symbol.kind() != SymbolKind.METHOD) {
+                NodeList<Token> qualifiers = functionDefinitionNode.qualifierList();
+                if (getQualifier(qualifiers, PUBLIC_KEYWORD) && !getQualifier(qualifiers, ISOLATED_KEYWORD)) {
+                    reportIssue(functionDefinitionNode, CoreRule.PUBLIC_NON_ISOLATED_CONSTRUCT);
+                }
+            }
+        });
+    }
+
+    private boolean getQualifier(List<Qualifier> qualifierList, String qualifierValue) {
```
poorna2152, line 172: Can't we simply do a `contains` check here. qualifierList.contains(qualifierValue)

scan-command/src/main/java/io/ballerina/scan/internal/StaticCodeAnalyzer.java
```diff
@@ -100,6 +116,77 @@ public void visit(ImplicitAnonymousFunctionExpressionNode implicitAnonymousFunct
         this.visitSyntaxNode(implicitAnonymousFunctionExpressionNode.expression());
     }
 
+    @Override
+    public void visit(ClassDefinitionNode classDefinitionNode) {
+        checkNonIsolatedPublicClassDefinitions(classDefinitionNode);
+        checkNonIsolatedPublicClassMembers(classDefinitionNode);
```
poorna2152, line 122: Can't we call this function with classDefinitionNode.members() directly.

scan-command/src/main/java/io/ballerina/scan/internal/StaticCodeAnalyzer.java
```diff
@@ -100,6 +116,77 @@ public void visit(ImplicitAnonymousFunctionExpressionNode implicitAnonymousFunct
...
+    }
+
+    private void checkNonIsolatedPublicClassDefinitions(ClassDefinitionNode classDefinitionNode) {
```
poorna2152, line 125: `checkNonIsolatedPublicClassDefinitions` -> `checkNonIsolatedPublicClassDefinition` to show that we are only checking one classdefinition inside the function. Similarly in other place as well

scan-command/src/main/java/io/ballerina/scan/internal/CoreRule.java
```diff
@@ -31,8 +31,9 @@
  * */
 enum CoreRule {
     AVOID_CHECKPANIC(RuleFactory.createRule(1, "Avoid checkpanic", RuleKind.CODE_SMELL)),
-    UNUSED_FUNCTION_PARAMETER(RuleFactory.createRule(2,
-            "Unused function parameter", RuleKind.CODE_SMELL));
+    UNUSED_FUNCTION_PARAMETER(RuleFactory.createRule(2, "Unused function parameter", RuleKind.CODE_SMELL)),
+    PUBLIC_NON_ISOLATED_CONSTRUCT(RuleFactory.createRule(3,
+            "Non isolated public class or function/method", RuleKind.CODE_SMELL));
```
MaryamZi, line 36: Can we break this down to two rules? Non-isolated object and non-isolated function/method.

scan-command/src/main/java/io/ballerina/scan/utils/Constants.java
```diff
@@ -61,6 +61,8 @@ public class Constants {
     static final String RULE_KIND_COLUMN = "Rule Kind";
     static final String RULE_DESCRIPTION_COLUMN = "Rule Description";
     static final String[] RULE_PRIORITY_LIST = {"ballerina", "ballerina/", "ballerinax/", "wso2/"};
+    public static final String MAIN_METHOD = "main";
+    public static final String INIT_METHOD = "init";
```
MaryamZi, line 65: Function, not method.

scan-command/src/main/java/io/ballerina/scan/internal/StaticCodeAnalyzer.java
MaryamZi, line 151: Aren't you repeating this bit unnecessarily foe each member?

scan-command/src/main/java/io/ballerina/scan/internal/StaticCodeAnalyzer.java
MaryamZi, line 140: This doesn't cover resources, does it?

scan-command/src/main/java/io/ballerina/scan/internal/StaticCodeAnalyzer.java
```diff
@@ -100,6 +116,77 @@ public void visit(ImplicitAnonymousFunctionExpressionNode implicitAnonymousFunct
...
+                member.accept(this);
+            });
+        });
+    }
+
+    private void checkNonIsolatedPublicMethods(FunctionDefinitionNode member,
+                                               ClassDefinitionNode classDefinitionNode) {
+        semanticModel.symbol(classDefinitionNode).ifPresent(symbol -> {
+            if (symbol instanceof ClassSymbol classSymbol) {
+                if (getQualifier(classSymbol.qualifiers(), PUBLIC_KEYWORD.stringValue()) &&
```
MaryamZi, line 152: Also repeated.

scan-command/src/test/resources/test-resources/core-rules/rule003_rule_non_isolated_public_functions.bal
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    function d() {
+
+    }
+
+    public function main() { // warning
+
+    }
+}
```
MaryamZi, line 101: Add tests for service, client, etc. classes and remote, resource methods.

scan-command/src/main/java/io/ballerina/scan/internal/StaticCodeAnalyzer.java
SasinduDilshara: For resources, the language itself gives a warning if it is not isolated.
So do we need to include it here also?

SasinduDilshara: Had a offline chat with @MaryamZi and decided that we don not need warnings for resources, instead of that add warnings for objects

SasinduDilshara: In there we use the value after casting. But here we just need to check the kind

scan-command/src/main/java/io/ballerina/scan/internal/StaticCodeAnalyzer.java
SasinduDilshara: Qualifier is a list of enum while qualifiervalue is a SyntaxKind

scan-command/src/main/java/io/ballerina/scan/internal/StaticCodeAnalyzer.java
SasinduDilshara: inside the `checkNonIsolatedPublicClassMembers` function we used classDefinition node itself.

scan-command/src/main/java/io/ballerina/scan/internal/StaticCodeAnalyzer.java
SasinduDilshara: Do you mean to directly cast the object without checking?

SasinduDilshara: I think it is biot clear handling those two things separately

scan-command/src/main/java/io/ballerina/scan/internal/StaticCodeAnalyzer.java
SasinduDilshara: I think it is bit clear handling those two things separately

scan-command/src/main/java/io/ballerina/scan/internal/CoreRule.java
MaryamZi: Still one rule?

scan-command/src/main/java/io/ballerina/scan/internal/StaticCodeAnalyzer.java
MaryamZi: Are you agreeing or? :)

My point is that you're doing this for each member, which isn't necessary since it doesn't change from member to member of the class.

Issue #29, opened 2024-10-31
Line numbers should exact same with the Ballerina file line numbers in Error tests in Static code analayzer


SasinduDilshara:
### Description

Please refer to https://github.com/ballerina-platform/static-code-analysis-tool/pull/24#discussion_r1812249627 for more details

### Steps to Reproduce

https://github.com/ballerina-platform/static-code-analysis-tool/pull/24#discussion_r1812249627

### Version

2201.10.2

gayaldassanayake:
## TLDR
**Coloring only the start line is the reason.**
Potential Options
1. We color from the start line to the end line - might look crowded
2. We go into a child node and put its location during the analysis
3. We leave as it is

## Explaination 
Debugged this and checked for the root cause. This issue arises when there is a line range given as the range of the issue.

For example,

``` ballerina
# Description.                               #L1
# + return - return value description        #L2
public function hello() returns string {     #L3
    return "Hello, World!";                  #L4
}                                            #L5
                                             #L6
``` 

The issue location is reported as follows.
``` json
{
    "location": {
        "filePath": "main.bal",
        "startLine": 0,
        "endLine": 4,
        "startColumn": 0,
        "endColumn": 1,
        "startOffset": 0,
        "length": 123
    }
}
```

<img width="688" alt="Image" src="https://github.com/user-attachments/assets/102b3b5e-838f-48e2-9679-55acf8ce33f9" />

In the above case, the node extends from line 0-4, However, we paint only the start line (L0). We need to discuss whether we are to paint all 5 lines of the function in this case. I believe painting only the start line was an active decision because painting all lines might make the report crowded.

Note - The Location class in Ballerina starts with index 0. However, in the HTML report, it begins with 1. This is corrected by adding 1 to the line number, so this is not the issue.

SasinduDilshara:
I believe we needs to highlight all the lines if the error occuers in few lines

gayaldassanayake:
In the sync-up it was discussed to,
1. Change the report to visualize line coverage from the first to last.
2. In rules like isolated functions, only use the location of the public keyword

gayaldassanayake:
This was discussed in the sync-up and here were the conclusions.
- Mark the entire startline->endline part as the issue area in the HTML report.
- In the isolated related issues, mark only the public keyword as the location.

PR #31, opened 2024-11-01
Fix incorrect error diagnostic location parameters

SasinduDilshara:
Fixes https://github.com/ballerina-platform/static-code-analysis-tool/issues/29

### DIFF
--- a/scan-command/src/main/java/io/ballerina/scan/internal/IssueImpl.java
+++ b/scan-command/src/main/java/io/ballerina/scan/internal/IssueImpl.java
@@ -45,9 +45,9 @@ public class IssueImpl implements Issue {
               String filePath) {
         LineRange lineRange = location.lineRange();
         TextRange textRange = location.textRange();
-        this.location = new BLangDiagnosticLocation(lineRange.fileName(), lineRange.startLine().line(),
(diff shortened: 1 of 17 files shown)


scan-command/src/main/java/io/ballerina/scan/internal/IssueImpl.java
```diff
@@ -45,9 +45,9 @@ public class IssueImpl implements Issue {
               String filePath) {
         LineRange lineRange = location.lineRange();
         TextRange textRange = location.textRange();
-        this.location = new BLangDiagnosticLocation(lineRange.fileName(), lineRange.startLine().line(),
-                lineRange.endLine().line(), lineRange.startLine().offset(), lineRange.endLine().offset(),
-                textRange.startOffset(), textRange.length());
+        this.location = new BLangDiagnosticLocation(lineRange.fileName(), lineRange.startLine().line() + 1,
+                lineRange.endLine().line() + 1, lineRange.startLine().offset() + 1, lineRange.endLine().offset() + 1,
+                textRange.startOffset() + 1, textRange.length());
```
MaryamZi, line 50: We shouldn't change this. We should change the assertion util method instead. Similar to `org.ballerinalang.test.BAssertUtil#validateError(org.ballerinalang.test.CompileResult, int, java.lang.String, int, int)`.

PR #37, opened 2025-02-19
Add static rule for Unused class fields

SasinduDilshara:
Rule Description - https://docs.google.com/document/d/117SJltpBJU4lb5hzhTZQe2rTOig_5O97tqetRzHeyU8/edit?tab=t.0#heading=h.pum8z720eanu


scan-command/src/main/java/io/ballerina/scan/internal/CoreRule.java
```diff
@@ -31,8 +31,8 @@
  * */
 enum CoreRule {
     AVOID_CHECKPANIC(RuleFactory.createRule(1, "Avoid checkpanic", RuleKind.CODE_SMELL)),
-    UNUSED_FUNCTION_PARAMETER(RuleFactory.createRule(2,
-            "Unused function parameter", RuleKind.CODE_SMELL));
+    UNUSED_FUNCTION_PARAMETER(RuleFactory.createRule(2, "Unused function parameter", RuleKind.CODE_SMELL)),
+    UNUSED_CLASS_FIELDS(RuleFactory.createRule(5, "Unused class private fields", RuleKind.CODE_SMELL));
```
MaryamZi, line 35:
```suggestion
    UNUSED_PRIVATE_CLASS_FIELD(RuleFactory.createRule(5, "Unused private class field", RuleKind.CODE_SMELL));
```

scan-command/src/test/java/io/ballerina/scan/internal/Rule005Test.java
MaryamZi, line 1: Licence header.

scan-command/src/test/java/io/ballerina/scan/internal/Rule005Test.java
```diff
@@ -0,0 +1,35 @@
...
+
+import io.ballerina.projects.Document;
+import io.ballerina.scan.Issue;
+import io.ballerina.scan.RuleKind;
+import org.testng.Assert;
+import org.testng.annotations.Test;
+
+import java.util.List;
+
+public class Rule005Test extends StaticCodeAnalyzerTest {
```
MaryamZi, line 11: Docs.

scan-command/src/test/java/io/ballerina/scan/internal/Rule005Test.java
```diff
@@ -0,0 +1,35 @@
...
+    private static final String UNUSED_CLASS_PRIVATE_FIELDS = "Unused class private fields";
+
+    @Test(description = "test checkpanic analyzer")
+    void testUnusedClassFieldsAnalyzer() {
```
MaryamZi, line 15: Needs to reflect "private".

scan-command/src/test/java/io/ballerina/scan/internal/Rule005Test.java
```diff
@@ -0,0 +1,35 @@
...
+import io.ballerina.scan.RuleKind;
+import org.testng.Assert;
+import org.testng.annotations.Test;
+
+import java.util.List;
+
+public class Rule005Test extends StaticCodeAnalyzerTest {
+    private static final String UNUSED_CLASS_PRIVATE_FIELDS = "Unused class private fields";
+
+    @Test(description = "test checkpanic analyzer")
```
MaryamZi, line 14: Wrong description. I would get rid of the description anyway, doesn't add value.

scan-command/src/test/java/io/ballerina/scan/internal/Rule005Test.java
```diff
@@ -0,0 +1,35 @@
...
+    void testUnusedClassFieldsAnalyzer() {
+        String documentName = "rule005_unused_class_fields.bal";
+        Document document = loadDocument(documentName);
+        ScannerContextImpl scannerContext = new ScannerContextImpl(List.of(CoreRule.UNUSED_CLASS_FIELDS.rule()));
+        StaticCodeAnalyzer staticCodeAnalyzer = new StaticCodeAnalyzer(document, scannerContext,
+                document.module().getCompilation().getSemanticModel());
+        staticCodeAnalyzer.analyze();
+        List<Issue> issues = scannerContext.getReporter().getIssues();
+        Assert.assertEquals(issues.size(), 5);
```
MaryamZi, line 23: Why not use a variable to assert total instead, like we do with compiler tests? That way we know we assert everything.

scan-command/src/test/resources/test-resources/core-rules/rule005_unused_class_fields.bal
```diff
@@ -0,0 +1,87 @@
...
+    }
+}
+
+service class SA {
+    private string b; // warning
+    private string c = "";
+    private string d = "";
+
+    function init() {
+        self.test(self.d);
```
MaryamZi, line 81: If we fix https://github.com/ballerina-platform/ballerina-lang/issues/23848, this test will fail. Shall we change the test to avoid that scenario?

MaryamZi (commented):
Title also needs to reflect that this is for private fields.

scan-command/src/test/java/io/ballerina/scan/internal/Rule005Test.java
SasinduDilshara: This will improve the readability and this will also checks the relevent fields in the rule as well

scan-command/src/test/java/io/ballerina/scan/internal/Rule005Test.java
SasinduDilshara: Will update

scan-command/src/test/java/io/ballerina/scan/internal/Rule005Test.java
SasinduDilshara: Using default qualifier is the way to write test functions

MaryamZi: That's not what I meant. I meant that it needs to say something like unused**Private**Fields.

scan-command/src/test/java/io/ballerina/scan/internal/Rule005Test.java
MaryamZi: > checks the relevent fields in the rule as well

Can you please elaborate?

Even in terms of readability, the number of errors doesn't really add value does it? If we follow the approach of using a variable that gets incremented with the assertion of each rule violation, we make sure that all the violations are asserted and consciously verified before making a change here too.

SasinduDilshara: Since this apply to all the rules, will create a separate issue to track

SasinduDilshara: Created https://github.com/ballerina-platform/static-code-analysis-tool/issues/47

keizer619 (approved):

Issue #42, opened 2025-02-20
The command for all rules printed in a invalid order


SasinduDilshara:
### Description

This is the order that prints for all issues

```
ballerina:1                                      | CODE_SMELL    | Avoid checkpanic
ballerina:10                                     | CODE_SMELL    | Invalid range operator
ballerina:2                                      | CODE_SMELL    | Unused function parameter
ballerina/example_module_static_code_analyzer:1  | CODE_SMELL    | rule 1
ballerina/example_module_static_code_analyzer:2  | BUG           | rule 2
ballerina/example_module_static_code_analyzer:3  | VULNERABILITY | rule 3
ballerinax/example_module_static_code_analyzer:1 | CODE_SMELL    | rule 1
ballerinax/example_module_static_code_analyzer:2 | BUG           | rule 2
ballerinax/example_module_static_code_analyzer:3 | VULNERABILITY | rule 3
exampleOrg/example_module_static_code_analyzer:1 | CODE_SMELL    | rule 1
exampleOrg/example_module_static_code_analyzer:2 | BUG           | rule 2
exampleOrg/example_module_static_code_analyzer:3 | VULNERABILITY | rule 3]
```

The order should be 1, 2, and 10

### Steps to Reproduce

$Subject

### Version

N/A

PR #43, opened 2025-02-20
Add Static rules for invalid range expression

SasinduDilshara:
Add static rule for https://docs.google.com/document/d/117SJltpBJU4lb5hzhTZQe2rTOig_5O97tqetRzHeyU8/edit?tab=t.0#heading=h.uefoxy928j7j


scan-command/src/main/java/io/ballerina/scan/internal/CoreRule.java
```diff
@@ -32,7 +32,8 @@
 enum CoreRule {
     AVOID_CHECKPANIC(RuleFactory.createRule(1, "Avoid checkpanic", RuleKind.CODE_SMELL)),
     UNUSED_FUNCTION_PARAMETER(RuleFactory.createRule(2,
-            "Unused function parameter", RuleKind.CODE_SMELL));
+            "Unused function parameter", RuleKind.CODE_SMELL)),
+    INVALID_RANGE_OPERATOR(RuleFactory.createRule(10, "Invalid range operator", RuleKind.CODE_SMELL));
```
MaryamZi, line 36: Shall we use expression instead of operator? Let's try and stick to spec terms wherever possible (where they also are clear to the user).

scan-command/src/main/java/io/ballerina/scan/internal/StaticCodeAnalyzer.java
```diff
@@ -72,6 +80,70 @@ public void visit(CheckExpressionNode checkExpressionNode) {
         }
     }
 
+    public void visit(BinaryExpressionNode binaryExpressionNode) {
+        if (binaryExpressionNode.operator().kind().equals(ELLIPSIS_TOKEN)
```
MaryamZi, line 84: Can extract `binaryExpressionNode.operator()` and then the kind to variables to avoid repeated method calls.

Please check and refactor the code across the PRs. Let's make sure we're following best practices.

scan-command/src/main/java/io/ballerina/scan/internal/StaticCodeAnalyzer.java
```diff
@@ -72,6 +80,70 @@ public void visit(CheckExpressionNode checkExpressionNode) {
...
+                || binaryExpressionNode.operator().kind().equals(DOUBLE_DOT_LT_TOKEN)) {
+            validateRangeExpressionOperator(scannerContext, document, binaryExpressionNode.lhsExpr(),
+                    binaryExpressionNode.rhsExpr(), binaryExpressionNode.operator(), binaryExpressionNode.location());
+        }
+    }
+
+    private void validateRangeExpressionOperator(ScannerContext scannerContext, Document document,
+                                                 Node n1, Node n2, Token operator, NodeLocation location) {
```
MaryamZi, line 92: Doesn't lhsNode and rhsNode make more sense here?

MaryamZi: And maintain consistency with lhsExpr and rhsExpr below.

scan-command/src/main/java/io/ballerina/scan/internal/StaticCodeAnalyzer.java
```diff
@@ -72,6 +80,70 @@ public void visit(CheckExpressionNode checkExpressionNode) {
...
+                lhsExpr = "-" + lhsExpr;
+            }
+        }
+
+        if (n2 instanceof BasicLiteralNode rhsExprNode) {
+            rhsExpr = rhsExprNode.literalToken().text();
+            if (isMinusOperatorPresentInRhs) {
+                rhsExpr = "-" + rhsExpr;
+            }
+        }
```
MaryamZi, line 123: Can write a method to do what you're doing to identify the string representation of a number and call it with both nodes instead?

scan-command/src/main/java/io/ballerina/scan/internal/StaticCodeAnalyzer.java
```diff
@@ -72,6 +80,70 @@ public void visit(CheckExpressionNode checkExpressionNode) {
...
+
+        if (lhsExpr != null && rhsExpr != null) {
```
MaryamZi, line 125: Check the inverse and return early.

scan-command/src/main/java/io/ballerina/scan/internal/StaticCodeAnalyzer.java
```diff
@@ -72,6 +80,70 @@ public void visit(CheckExpressionNode checkExpressionNode) {
...
+            // According to the spec, these literal tokens are integers. So no need to cast or check.
+            try {
+                int lhs = Integer.parseInt(lhsExpr);
+                int rhs = Integer.parseInt(rhsExpr);
+                if (operator.kind() == DOUBLE_DOT_LT_TOKEN) {
+                    if (lhs >= rhs) {
+                        scannerContext.getReporter().reportIssue(document, location,
+                                CoreRule.INVALID_RANGE_OPERATOR.rule());
+                    }
+                } else if (operator.kind() == ELLIPSIS_TOKEN) {
```
MaryamZi, line 135: Isn't this always true? Can be an else (can also use early returns then).

scan-command/src/test/java/io/ballerina/scan/internal/Rule006Test.java
```diff
@@ -0,0 +1,66 @@
...
+import io.ballerina.projects.Document;
+import io.ballerina.scan.Issue;
+import io.ballerina.scan.RuleKind;
+import org.testng.Assert;
+import org.testng.annotations.Test;
+
+import java.util.List;
+
+/**
+ * Unused function parameters analyzer tests.
```
MaryamZi, line 30: Wrong description.

scan-command/src/test/resources/command-outputs/unix/print-rules-to-console.txt
```diff
@@ -2,6 +2,7 @@ Loading scan tool configurations from src/test/resources/test-resources/bal-proj
 	RuleID                                           | Rule Kind     | Rule Description         
 	---------------------------------------------------------------------------------------------
 	ballerina:1                                      | CODE_SMELL    | Avoid checkpanic
+	ballerina:10                                     | CODE_SMELL    | Invalid range operator
```
MaryamZi, line 5: This is wrong, isn't it? 10 should come after 2.

scan-command/src/test/resources/test-resources/core-rules/rule006_invalid_range_expression.bal
MaryamZi, line 1: Please make sure licence headers are added.

scan-command/src/test/resources/test-resources/core-rules/rule006_invalid_range_expression.bal
```diff
@@ -0,0 +1,65 @@
...
+        // code
+    }
+
+    foreach int i in 10..<10 { // warning
+        // code
+    }
+
+    foreach int i in -2..<-2 { // warning
+        // code
+    }
```
MaryamZi, line 64: Add a test or two with the range expressions without foreach loops. They are just another kind of expression.

MaryamZi (commented):
We can probably do something like this for `int:range` also. Can be a separate rule.

SasinduDilshara:
> We can probably do something like this for `int:range` also. Can be a separate rule.

Will create a issue and add that later

scan-command/src/main/java/io/ballerina/scan/internal/StaticCodeAnalyzer.java
SasinduDilshara: I guess it is better to use else if and show the condition.
It will Improve the readability

scan-command/src/test/resources/command-outputs/unix/print-rules-to-console.txt
SasinduDilshara: Yes, there is an already a issue.
It will fix on that https://github.com/ballerina-platform/static-code-analysis-tool/issues/42

keizer619 (approved):

Issue #66, opened 2025-06-03
Refactor scan tool to retrieve rule metadata directly from `Rule` instance instead of `rules.json`


nureka-rodrigo:
### Current Limitation

The current implementation of the Ballerina scan tool relies on a `rules.json` file to store and retrieve metadata such as rule IDs, descriptions, and kinds. Currently the usage of `reportIssue(...)` method from standard library compiler-plugins utilizes only a numeric rule ID to identify issues, with all descriptive metadata being pulled from `rules.json`. 

### Suggested Improvement

Refactor the scan tool to retrieve rule metadata directly from the passed `Rule` instance rather than relying on the `rules.json` file.

PR #67, opened 2025-06-03
Refactor the rule loading mechanism to use `ServiceLoader` Pattern

nureka-rodrigo:
## Purpose

Fixes https://github.com/ballerina-platform/static-code-analysis-tool/issues/66

## Approach

Replaced the `JSON`-based rule loading mechanism with Java's `ServiceLoader` pattern. Before the rules were loaded by parsing a `rules.json` file from the plugin `JAR` using `Gson`. Rules are now loaded using `ServiceLoader<RuleProvider>`, where external plugins implement the `RuleProvider` interface.

## Limitation

1. The scan tool is no longer backward compatible with old compiler plugins. Existing plugins will need to be updated to work with this new approach.

## Migration Guide for Compiler Plugin Authors

With this change, external compiler plugins need to be updated to use the new `ServiceLoader`-based approach:

1. Rules should now be defined as enum classes that implement the `Rule` interface:
```java
public enum OSRule implements Rule {
    AVOID_UNSANITIZED_CMD_ARGS(1, "Avoid constructing system command arguments from user " +
            "input without proper sanitization", VULNERABILITY),
    AVOID_UNSANITIZED_ENV_VARS(2, "Environment variables should not be defined from untrusted input",
            VULNERABILITY);

    private final int id;
    private final String description;
    private final RuleKind kind;

    OSRule(int id, String description, RuleKind kind) {
        this.id = id;
        this.description = description;
        this.kind = kind;
    }

    @Override
    public String id() {
        return "";
    }

    @Override
    public int numericId() {
        return id;
    }

    @Override
    public String description() {
        return description;
    }

    @Override
    public RuleKind kind() {
        return kind;
    }
}
```
2. The compiler plugin code analyzer should implement the `RuleProvider` with a **public default constructor**:
```java
public class OSStaticCodeAnalyzer extends CodeAnalyzer implements RuleProvider {
    private Reporter reporter;

    public OSStaticCodeAnalyzer() {
        // Default constructor for service loading
    }

    public OSStaticCodeAnalyzer(Reporter reporter) {
        this.reporter = reporter;
    }

    @Override
    public void init(CodeAnalysisContext analysisContext) {
        analysisContext.addSyntaxNodeAnalysisTask(new OSCommandInjectionAnalyzer(reporter), List.of(FUNCTION_CALL));
    }

    @Override
    public Iterable<Rule> getRules() {
        return List.of(OSRule.values());
    }
}
```

3. The Service should be registered in `META-INF/services/io.ballerina.scan.RuleProvider` in the compiler plugin JAR:
```java
io.ballerina.stdlib.os.compiler.staticcodeanalyzer.OSStaticCodeAnalyzer
```

4. The `rules.json` file is no longer needed and can be removed from the plugin resources.

## Check List

- [ ] Read the [Contributing Guide](https://github.com/ballerina-platform/ballerina-lang/blob/master/CONTRIBUTING.md)
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Integration Tests


scan-command/src/main/java/io/ballerina/scan/internal/ProjectAnalyzer.java
```diff
@@ -144,22 +116,15 @@ Map<String, List<Rule>> getExternalAnalyzers() {
...
                         return;
                     }
                     CompilerPluginDescriptor pluginDesc = pkgManifest.compilerPluginDescriptor().get();
-                    Optional<String> ruleFileContent = loadRuleFileContent(pluginName, pluginDesc);
-                    if (ruleFileContent.isEmpty()) {
-                        return;
-                    }
-                    List<Rule> externalRules = loadExternalRules(org, name, pluginName, ruleFileContent.get());
-                    externalAnalyzers.put(pluginDesc.plugin().getClassName(), externalRules);
+                    externalAnalyzers.put(pluginDesc.plugin().getClassName(), new ArrayList<>());
```
MaryamZi, line 127: Where are the external rules loaded now?

scan-command/src/main/java/io/ballerina/scan/internal/ReporterImpl.java
```diff
@@ -65,7 +65,19 @@ public void reportIssue(Document reportedDocument, Location location, int ruleId
 
     @Override
     public void reportIssue(Document reportedDocument, Location location, Rule rule) {
-        issues.add(createIssue(reportedDocument, location, rule));
+        if (rule instanceof RuleImpl) {
+            issues.add(createIssue(reportedDocument, location, rule));
+            return;
+        }
+
+        RuleImpl proxyRule = new RuleImpl(
```
MaryamZi, line 73: Why do we need this? Why doesn't L69 work for all? That accepts Rule, right?

MaryamZi: My bad, reading the description, ignore for now please.

nureka-rodrigo: My bad, the proxy pattern is unnecessary. `Instanceof` check is enough. 

Fix be04b5867d3986cbf30005fe2d37e22dd96693e8

scan-command/src/main/java/io/ballerina/scan/internal/ReporterImpl.java
```diff
@@ -65,7 +65,9 @@ public void reportIssue(Document reportedDocument, Location location, int ruleId
 
     @Override
     public void reportIssue(Document reportedDocument, Location location, Rule rule) {
-        issues.add(createIssue(reportedDocument, location, rule));
+        if (rule instanceof RuleImpl) {
```
MaryamZi, line 68: But this won't work for custom implementations.

scan-command/src/main/java/io/ballerina/scan/internal/ProjectAnalyzer.java
```diff
@@ -220,47 +167,75 @@ private InputStream loadResource(CompilerPluginDescriptor pluginDesc) {
...
-        if (!element.isJsonArray()) {
-            throw new ScanToolException(DiagnosticLog.error(DiagnosticCode.INVALID_JSON_FORMAT, RULES_FILE,
-                    pluginName, gson.toJson(element)));
+    private Class<?> findRuleConstantsClass(URLClassLoader classLoader, String pluginClassName) {
+        try {
+            String[] parts = pluginClassName.split("\\.");
+            if (parts.length >= 4) {
+                String stdlibName = parts[3];
+                String stdlibPath = "io.ballerina.stdlib." + stdlibName + ".compiler.staticcodeanalyzer."
+                        + RULE_CONSTANTS_CLASS;
```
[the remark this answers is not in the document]
nureka-rodrigo: a65984f4be299ec0c33ce0c47d6a458683e2be20

scan-command/src/main/java/io/ballerina/scan/internal/ReporterImpl.java
nureka-rodrigo: a65984f4be299ec0c33ce0c47d6a458683e2be20

PR #69, opened 2025-06-23
Refactor the rule loading mechanism to use ServiceLoader Pattern

### LINKED ISSUE #66: Refactor scan tool to retrieve rule metadata directly from `Rule` instance instead of `rules.json`

nureka-rodrigo:
## Purpose

Fixes https://github.com/ballerina-platform/static-code-analysis-tool/issues/66

## Approach

Replaced the `JSON`-based rule loading mechanism with Java's `ServiceLoader` pattern. Before the rules were loaded by parsing a `rules.json` file from the plugin `JAR` using `Gson`. Rules are now loaded using `ServiceLoader<RuleProvider>`, where external plugins implement the `RuleProvider` interface.

> [!CAUTION]
> The scan tool is no longer backward compatible with old compiler plugins. Existing plugins will need to be updated to work with this new approach. 

## Migration Guide for Compiler Plugin Authors

With this change, external compiler plugins need to be updated to use the new `ServiceLoader`-based approach:

1. Rules should now be defined as `Enum` classes that implement the `Rule` interface (see an example [implementation](https://github.com/nureka-rodrigo/module-ballerina-os/blob/test/compiler-plugin/src/main/java/io/ballerina/stdlib/os/compiler/staticcodeanalyzer/OSRule.java)):
```java
public enum OSRule implements Rule {
    AVOID_UNSANITIZED_CMD_ARGS(1, "Avoid constructing system command arguments from user " +
            "input without proper sanitization", VULNERABILITY),
    AVOID_UNSANITIZED_ENV_VARS(2, "Environment variables should not be defined from untrusted input",
            VULNERABILITY);

    private final int id;
    private final String description;
    private final RuleKind kind;

    OSRule(int id, String description, RuleKind kind) {
        this.id = id;
        this.description = description;
        this.kind = kind;
    }

    @Override
    public String id() {
        return "";
    }

    @Override
    public int numericId() {
        return id;
    }

    @Override
    public String description() {
        return description;
    }

    @Override
    public RuleKind kind() {
        return kind;
    }
}
```
2. A custom rule provider class that implement `RuleProvider` interface should be created (see an example [implementation](https://github.com/nureka-rodrigo/module-ballerina-os/blob/test/compiler-plugin/src/main/java/io/ballerina/stdlib/os/compiler/staticcodeanalyzer/OSRuleProvider.java)):
```java
public class OSRuleProvider implements RuleProvider {  
    @Override  
  public Iterable<Rule> getRules() {  
        return List.of(OSRule.values());  
    }  
}
```

3. The Service should be registered in `META-INF/services/io.ballerina.scan.RuleProvider` in the compiler plugin JAR (see an example [implementation](https://github.com/nureka-rodrigo/module-ballerina-os/blob/test/compiler-plugin/src/main/resources/META-INF/services/io.ballerina.scan.RuleProvider)):
```java
io.ballerina.stdlib.os.compiler.staticcodeanalyzer.OSRuleProvider
```

4. The `rules.json` file is no longer needed and can be removed from the plugin resources.

## Check List

- [x] Read the [Contributing Guide](https://github.com/ballerina-platform/ballerina-lang/blob/master/CONTRIBUTING.md)
- [ ] Added necessary tests
  - [x] Unit Tests
  - [ ] Integration Tests


nureka-rodrigo:
@MaryamZi @gayaldassanayake

test-compiler-plugins/ballerinax-plugin/src/main/java/io/ballerinax/examplestaticcodeanalyzer/CustomRule.java
```diff
@@ -0,0 +1,45 @@
...
+import io.ballerina.scan.RuleKind;
+
+import static io.ballerina.scan.RuleKind.BUG;
+import static io.ballerina.scan.RuleKind.CODE_SMELL;
+import static io.ballerina.scan.RuleKind.VULNERABILITY;
+
+public enum CustomRule implements Rule {
+    RULE_1(1, "rule 1", CODE_SMELL),
+    RULE_2(2, "rule 2",
+            BUG),
```
keizer619, line 13:
```suggestion
    RULE_2(2, "rule 2", BUG),
```

test-compiler-plugins/ballerina-plugin/src/main/java/io/ballerina/examplestaticcodeanalyzer/CustomRule.java
```diff
@@ -0,0 +1,45 @@
+package io.ballerina.examplestaticcodeanalyzer;
```
keizer619, line 1: Usually we have multiple levels for package name like this `io.ballerina.runtime.observability.metrics`
better to rename this

PR #76, opened 2025-09-25
Move "running scans" output message earlier in the execution flow

### LINKED ISSUE #75: The "Running Scans" message takes a while to appear after executing the command
### Current Limitation

Right now, it takes a few seconds for the "Running Scans" message to appear. It feels like the command is hanging.

https://github.com/user-attachments/assets/96c752bf-6fe9-41de-87de-ecb5ab339a04

### Suggested Improvement

The tool should be improved to show the "Running Scans" message immediately when the command is executed. 

### Version

At least 0.11.0

nureka-rodrigo:
## Purpose

Fixes: https://github.com/ballerina-platform/static-code-analysis-tool/issues/75

In the original implementation, it took a considerable amount of time for the 'Running scans' text to appear in the console, because several time-consuming operations were performed beforehand.

1. Loading project configurations
2. Getting external analyzers
3. Loading platform plugins
4. Processing rules to include/exclude

## Approach

This change moved the "Running Scans" message much earlier in the execution flow, right after the basic validation but before the heavy operations.

image

## Check List

- [ ] Read the [Contributing Guide](https://github.com/ballerina-platform/ballerina-lang/blob/master/CONTRIBUTING.md)
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Integration Tests

### DIFF
--- a/scan-command/src/main/java/io/ballerina/scan/internal/ScanCmd.java
+++ b/scan-command/src/main/java/io/ballerina/scan/internal/ScanCmd.java
@@ -57,7 +57,7 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.ServiceLoader;
 
-import static io.ballerina.scan.internal.ScanToolConstants.RUNNING_SCANS;
+import static io.ballerina.scan.internal.ScanToolConstants.RUNNING_SCANS_LOG;
 import static io.ballerina.scan.internal.ScanToolConstants.SCAN_COMMAND;
 
 /**
@@ -187,7 +187,7 @@ public class ScanCmd implements BLauncherCmd {
         }
 
         outputStream.println();
-        outputStream.println(RUNNING_SCANS);
+        outputStream.println(RUNNING_SCANS_LOG);
 
         ProjectAnalyzer projectAnalyzer = getProjectAnalyzer(project.get(), scanTomlFile.get());
         List<Rule> coreRules = CoreRule.rules();
--- a/scan-command/src/main/java/io/ballerina/scan/internal/ScanCmd.java
+++ b/scan-command/src/main/java/io/ballerina/scan/internal/ScanCmd.java
@@ -57,6 +57,7 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.ServiceLoader;
 
+import static io.ballerina.scan.internal.ScanToolConstants.RUNNING_SCANS;
 import static io.ballerina.scan.internal.ScanToolConstants.SCAN_COMMAND;
 
 /**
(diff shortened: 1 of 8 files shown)


nureka-rodrigo:
@azinneera @keizer619 PRAM

scan-command/src/main/java/io/ballerina/scan/internal/ScanCmd.java
```diff
@@ -185,6 +185,9 @@ public void execute() {
             return;
         }
 
+        outputStream.println();
+        outputStream.println("Running Scans");
```
SasinduDilshara:
```suggestion
        outputStream.println("\nRunning Scans");
```

Better we can add this as a constant variable and use

nureka-rodrigo: Fix: 801dbf929204ea4f9a70c1d05fba7f03ba6d08e1

nureka-rodrigo: @SasinduDilshara I think using `println()` instead of `\n` is the better approach. The `println()` method is platform-independent and explicitly terminates the line using the system’s line separator, which is safer than hardcoding `\n`. As the official Java [PrintStream](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/io/PrintStream.html) documentation states, 

>_Terminates the current line by writing the line separator string. The line separator string is defined by the system property `line.separator`, and is not necessarily a single newline character ('\n')._ 

While `\n` works on Unix-like systems, it may not represent the correct line ending on Windows or other environments. Using `println()` also makes the intent to end the line clearer to readers and avoids subtle inconsistencies in cross-platform code, which is why I think it is the better option here.

scan-command/src/main/java/io/ballerina/scan/internal/ScanToolConstants.java
```diff
@@ -38,6 +38,7 @@ public class ScanToolConstants {
     static final String CODE_SMELL = "CODE_SMELL";
     static final String RULE_ID = "id";
     static final String RULE_DESCRIPTION = "description";
+    static final String RUNNING_SCANS = "Running Scans";
```
gayaldassanayake:
```suggestion
    static final String RUNNING_SCANS_LOG = "Running Scans";
```

nureka-rodrigo: Fix: daa5516ab2d7fb5877ef452b99cb4239c408cdb2

keizer619 (approved):

